PR #2, opened 2020-11-10
Add assertion macros

Phantomical:
These are mostly just the same as we had in decaf but they've been renamed from `DECAF_` to `CAFFEINE_`


mishazharov (approved):
Nice

PR #3, opened 2020-11-10
Add vcpkg.json and build-project github action

Phantomical:
This adds a simple github action that uses the existing docker image to build + test everything.


vcpkg.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "decaf",
```
mishazharov, line 2: Caffeine

mishazharov (approved):
Otherwise lgtm

PR #5, opened 2020-11-10
Setup unit test runner

Phantomical:
No tests at the moment but when they get added they can now be run automatically.


Skrinikov (approved):
LGTM

PR #14, opened 2020-11-16
Implement solver API

Phantomical:
This implements a fairly simple solver interface that has two main methods:
- `check`: Check whether the assertions are satisfiable and return `SAT`, `UNSAT`, or `Unknown` as applicable. This should be faster if a model isn't needed.
- `resolve`: Check whether the assertions are satisfiable and return a model that has the necessary assigned symbolic values.

It also adds an `Assertion` type (which is an expression that is always a boolean) and a new `Constant` expression that represents a symbolic constant.


mishazharov (approved):
LGTM

PR #17, opened 2020-11-19
Add formatter & run format

mishazharov:
This PR adds the format cmake target and the github workflow for linting

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,3 +22,18 @@ include(setup-warnings)
 
 add_subdirectory(src)
 add_subdirectory(test)
+
+file(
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -22,3 +22,18 @@ include(setup-warnings)
 
 add_subdirectory(src)
 add_subdirectory(test)
+
+file(
+  GLOB_RECURSE fmt_sources
+  *.c
+  *.cpp
+  *.h
+  *.hpp
```
Phantomical, line 31: You should probably add `*.inl` here as well.

Phantomical (approved):
LGTM as long as the comment is addressed.

PR #18, opened 2020-11-19
Add operation visitor

Phantomical:
This is similar to LLVM's InstVisitor but for operations. This PR also adds hashing support to Type and Operation.


Skrinikov (approved):
looks good.

PR #19, opened 2020-11-20
Add z3 solver struture

Skrinikov:
This is a WIP PR to see whether this is on the right track

### DIFF
--- a/include/caffeine/Solver/Z3Solver.h
+++ b/include/caffeine/Solver/Z3Solver.h
@@ -4,11 +4,11 @@
 #include <memory>
 #include <z3++.h>
 
-#include "caffeine/IR/Operation.h"
 #include "caffeine/IR/Assertion.h"
+#include "caffeine/IR/Operation.h"
 #include "caffeine/IR/Value.h"
-#include "caffeine/Solver/Solver.h"
 #include "caffeine/IR/Visitor.h"
+#include "caffeine/Solver/Solver.h"
 
 namespace caffeine {
 
@@ -57,10 +57,11 @@ class Z3Model : public Model {
 protected:
   z3::context* ctx;
   z3::model model;
-  std::map<std::string, z3::expr*> constants; 
+  std::map<std::string, z3::expr*> constants;
 
 public:
-  Z3Model(SolverResult, z3::context*, z3::model, std::map<std::string, z3::expr*>);
+  Z3Model(SolverResult, z3::context*, z3::model,
+          std::map<std::string, z3::expr*>);
   /**
    * Evaluate an expression using this model. Returns an appropriate constant
    * expression (i.e. is_constant returns true) with the value of said constant.
--- a/include/caffeine/Solver/Z3Solver.h
+++ b/include/caffeine/Solver/Z3Solver.h
@@ -47,14 +47,10 @@ public:
   // Unary operations
   z3::expr visitNot (const UnaryOp& op);
   z3::expr visitFNeg(const UnaryOp& op);
-
   // clang-format on
-protected:
-  z3::context* ctx;
-  z3::solver* solver;
 };
 
-class Z3Model : Model {
+class Z3Model : public Model {
 protected:
   z3::context* ctx;
   z3::model model;
--- a/include/caffeine/Solver/Z3Solver.h
+++ b/include/caffeine/Solver/Z3Solver.h
@@ -5,6 +5,8 @@
 #include <z3++.h>
 
 #include "caffeine/IR/Operation.h"
+#include "caffeine/IR/Assertion.h"
+#include "caffeine/IR/Value.h"
 #include "caffeine/Solver/Solver.h"
 #include "caffeine/IR/Visitor.h"
 
@@ -12,9 +14,10 @@ namespace caffeine {
 
 class Z3OpVisitor : public ConstOpVisitor<Z3OpVisitor, z3::expr> {
   z3::context* ctx;
+  std::map<std::string, z3::expr*>& constMap;
 
 public:
-  Z3OpVisitor(z3::context* ctx);
+  Z3OpVisitor(z3::context* ctx, std::map<std::string, z3::expr*>& constMap);
 
   // clang-format off
   z3::expr visitConstant     (const Constant& op);
@@ -54,16 +57,17 @@ class Z3Model : public Model {
 protected:
   z3::context* ctx;
   z3::model model;
+  std::map<std::string, z3::expr*> constants; 
 
 public:
-  Z3Model(SolverResult, z3::context*, z3::model);
+  Z3Model(SolverResult, z3::context*, z3::model, std::map<std::string, z3::expr*>);
   /**
    * Evaluate an expression using this model. Returns an appropriate constant
    * expression (i.e. is_constant returns true) with the value of said constant.
    *
    * It is invalid to call this method if the model is not SAT.
    */
-  ref<Operation> evaluate(const ref<Operation>& expr) const;
+  Value evaluate(const ref<Operation>& expr) const;
 
   /**
    * Look up the value of a symbolic constant in this model. Returns an
@@ -73,7 +77,7 @@ public:
    *
    * It is invalid to call this method if the model is not SAT.
    */
-  ref<Operation> lookup(const Constant& constant) const;
+  Value lookup(const Constant& constant) const;
 }; // class Z3Model
 
 class Z3Solver : public Solver {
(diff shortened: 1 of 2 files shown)


include/caffeine/Solver/Z3Solver.h
```diff
@@ -0,0 +1,139 @@
+#ifndef CAFFEINE_SOLVER_Z3_H
+#define CAFFEINE_SOLVER_Z3_H
+
+#include <memory>
+#include <z3++.h>
+
+#include "Solver.h"
```
Phantomical, line 7: Nit: this should probably be 
```
#include "caffeine/IR/Solver.h"
```
to be consistent.

include/caffeine/Solver/Z3Solver.h
```diff
@@ -0,0 +1,139 @@
...
+#include "caffeine/IR/Operation.h"
+#include "caffeine/IR/Visitor.h"
+
+namespace caffeine {
+
+class Z3OpVisitor : public ConstOpVisitor<Z3OpVisitor, z3::expr> {
+  Z3OpVisitor(z3::context *ctx, z3::solver *solver);
+
+  // clang-format off
+  z3::expr visitConstant     (Constant& op);
```
Phantomical, line 17: Unless I've messed something up, for `ConstOpVisitor` these should take const references (`const Constant& op`).

I suspect you'd find that once you try to run this none of the visit methods would be called.

include/caffeine/Solver/Z3Solver.h
```diff
@@ -0,0 +1,139 @@
...
+
+public:
+  Z3Model(SolverResult, z3::context*, z3::model);
+  /**
+   * Evaluate an expression using this model. Returns an appropriate constant
+   * expression (i.e. is_constant returns true) with the value of said constant.
+   *
+   * It is invalid to call this method if the model is not SAT.
+   */
+  ref<Operation> evaluate(const ref<Operation>& expr) const;
```
Phantomical, line 67: I should have a PR up tonight that implements a general `evaluate` so you can probably avoid implementing this for now.

src/Solver/Z3Solver.cpp
```diff
@@ -0,0 +1,81 @@
...
+    // in the future we can get an API for doing this more efficiently
+    // added to Z3.
+    llvm::SmallString<64> str;
+    op.value().toStringUnsigned(str, 10);
+
+    return ctx->bv_val(str.c_str(), op.value().getBitWidth());
+}
+
+z3::expr Z3OpVisitor::visitConstantFloat(ConstantFloat& op) {
+  return ctx->fpa_val(op.value().convertToFloat()); // TODO: Is this corret?
```
Phantomical, line 59: We'll have to figure out a more general way to do this. It should be good enough for now but when we merge this we should open an issue to resolve it.

Phantomical (commented):
It looks like you're on the right track :) Left a few comments on things that looked fishy but that's it.

src/Solver/Z3Solver.cpp
```diff
@@ -0,0 +1,169 @@
...
+CAFFEINE_OP_IMPL_DELEGATE(SDiv, lhs / rhs)
+CAFFEINE_OP_IMPL_DELEGATE(URem, z3::urem(lhs, rhs))
+CAFFEINE_OP_IMPL_DELEGATE(SRem, lhs % rhs)
+CAFFEINE_OP_IMPL_DELEGATE(And, lhs & rhs)
+CAFFEINE_OP_IMPL_DELEGATE(Or, lhs | rhs)
+CAFFEINE_OP_IMPL_DELEGATE(Xor, lhs ^ rhs)
+CAFFEINE_OP_IMPL_DELEGATE(Shl, z3::shl(lhs, rhs))
+CAFFEINE_OP_IMPL_DELEGATE(LShr, z3::lshr(lhs, rhs))
+CAFFEINE_OP_IMPL_DELEGATE(AShr, z3::ashr(lhs, rhs))
+CAFFEINE_OP_IMPL_DELEGATE(FAdd, lhs + rhs) // I did not see specific functions for floats?
```
Skrinikov, line 83: At least in z3

src/Solver/Z3Solver.cpp
```diff
@@ -0,0 +1,255 @@
...
+  case ICmpOpcode::SGT:
+    return lhs > rhs;
+  case ICmpOpcode::SGE:
+    return lhs >= rhs;
+  case ICmpOpcode::SLT:
+    return lhs < rhs;
+  case ICmpOpcode::SLE:
+    return lhs <= rhs;
+  default:
+    CAFFEINE_ASSERT(false, "Unknown ICmpOpcode");
```
Phantomical, line 203: nit: You can just use `CAFFEINE_ABORT` here

src/Solver/Z3Solver.cpp
```diff
@@ -0,0 +1,255 @@
...
+#include "caffeine/IR/Type.h"
+#include "caffeine/Support/Assert.h"
+
+namespace caffeine {
+
+Z3Model::Z3Model(SolverResult result, z3::context* ctx, z3::model model,
+                 std::map<std::string, z3::expr*> map)
+    : Model(result), ctx(ctx), model(model), constants(map) {}
+
+Value Z3Model::evaluate(const ref<Operation>& expr) const {
```
Phantomical, line 11: You don't need to implement this. It's already implemented for the base `Model` class (and what you have here is wrong since it doesn't evaluate expression trees).

include/caffeine/Solver/Z3Solver.h
```diff
@@ -0,0 +1,143 @@
...
+#include "caffeine/IR/Operation.h"
+#include "caffeine/IR/Value.h"
+#include "caffeine/IR/Visitor.h"
+#include "caffeine/Solver/Solver.h"
+
+namespace caffeine {
+
+class Z3OpVisitor : public ConstOpVisitor<Z3OpVisitor, z3::expr> {
+  z3::context* ctx;
+  std::map<std::string, z3::expr*>& constMap;
```
Phantomical, line 17: Z3 expressions are reference counted so you can avoid the pointer here.

src/Solver/Z3Solver.cpp
```diff
@@ -0,0 +1,255 @@
...
+
+  std::map<std::string, z3::expr*> cc;
+  Z3OpVisitor visitor(ctx, cc);
+  auto expression = visitor.visit(expr.get());
+
+  auto result = model.eval(expression, false);
+
+  if (result == z3::sat) {
+    if (result.is_int()) {
+      return Value(llvm::APInt(32, result.get_numeral_int()));
```
Phantomical, line 22: Use the bitwidth of the type here (either from the z3 sort or from the constant type)

src/Solver/Z3Solver.cpp
```diff
@@ -0,0 +1,255 @@
...
+Value Z3Model::lookup(const Constant& constant) const {
+  CAFFEINE_ASSERT(result() == SolverResult::SAT, "Model is not SAT");
+
+  auto it = constants.find(std::string(constant.name()));
+  if (it == constants.end()) {
+    return Value();
+  }
+
+  if (it->second->is_int()) {
+    return Value(llvm::APInt(32, it->second->get_numeral_int()));
```
Phantomical, line 40: Same comment about bitwidth here.

src/Solver/Z3Solver.cpp
```diff
@@ -0,0 +1,255 @@
...
+  Z3OpVisitor visitor(ctx, cc);
+  auto expression = visitor.visit(expr.get());
+
+  auto result = model.eval(expression, false);
+
+  if (result == z3::sat) {
+    if (result.is_int()) {
+      return Value(llvm::APInt(32, result.get_numeral_int()));
+    } else {
+      return Value(); // I cant find the way to extract fpa
```
Phantomical, line 24: Yeah we can solve this in a follow-up PR.

include/caffeine/Solver/Z3Solver.h
```diff
@@ -0,0 +1,143 @@
...
+  z3::expr visitNot (const UnaryOp& op);
+  z3::expr visitFNeg(const UnaryOp& op);
+  // clang-format on
+};
+
+class Z3Model : public Model {
+protected:
+  z3::context* ctx;
+  z3::model model;
+  std::map<std::string, z3::expr*> constants;
```
Phantomical, line 60: Same note here WRT to `z3::expr` and pointers.

Phantomical (changes requested):
There's a few easily fixed but major problems. I've marked those.

Anything else we run into can be fixed later.

Phantomical (approved):

PR #20, opened 2020-11-20
Add default implementation of Model::evaluate

Phantomical:
The logic to actually implement this is decently complicated (it requires walking the expression tree) an is entirely implementable using only the `Model::lookup` function. This PR implements `Model::evaluate` in terms of `Model::lookup`. Note that the implementation is still incomplete, I'll actually fill everything out in a follow-up PR.

To do this I found I needed to introduce a new `Value` as having the representation be an operation that's constrained to be either `ConstantInt` or `ConstantFloat` ended up being basically unusable for actually performing arithmetic. I've also changed the model interface to use `Value` as that made far more sense.

While doing this I went and added some utility methods to `Operation` and `Type` as I discovered that I needed them. As a result, there's now a method to get the opcode name as a string, a shortcut for creating boolean types, and operation operands can now be accessed through the indexing operator.

@Skrinikov This PR changes the model interface slightly so it's going to require some minor changes to your PR.

### DIFF
--- a/src/IR/Value.cpp
+++ b/src/IR/Value.cpp
@@ -1,0 +1,277 @@
+#include "caffeine/IR/Value.h"
+#include "caffeine/IR/Operation.h"
+#include "caffeine/IR/Type.h"
+#include "caffeine/Support/Assert.h"
+
+namespace caffeine {
+
+Value::Value() : kind_(Empty) {}
+
+Value::Value(const llvm::APInt& apint) : kind_(Int), apint_(apint) {}
+Value::Value(llvm::APInt&& apint) : kind_(Int), apint_(std::move(apint)) {}
+
+Value::Value(const llvm::APFloat& apfloat) : kind_(Float), apfloat_(apfloat) {}
+Value::Value(llvm::APFloat&& apfloat) : kind_(Float), apfloat_(apfloat) {}
+
+Value::Value(const ConstantInt& constant) : Value(constant.value()) {}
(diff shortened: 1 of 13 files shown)


src/IR/Value.cpp
```diff
@@ -0,0 +1,277 @@
...
+}
+
+Value Value::select(const Value& cond, const Value& t, const Value& f) {
+  CAFFEINE_ASSERT(cond.type() == Type::bool_ty());
+  CAFFEINE_ASSERT(t.type() == f.type());
+
+  return cond.apint() == 1 ? t : f;
+}
+
+} // namespace caffeine
```
Skrinikov, line 277: fmt :)

src/Solver/Solver.cpp
```diff
@@ -1,10 +1,76 @@
...
+    return model_->lookup(op);
+  }
+  Value visitConstantInt(const ConstantInt& op) {
+    return op.value();
+  }
+  Value visitConstantFloat(const ConstantFloat& op) {
+    return op.value();
+  }
+
+#define DECL_BINOP(opcode, func)                                               \
```
Skrinikov, line 28: Maybe worth undefining after use?

Phantomical: I can if we think it's important. It's in a cpp file though so I don't think it really matters? (header files are a different situation).

Skrinikov (approved):
LGTM

Issue #7, opened 2020-11-16
Implement interpreter


mishazharov:
Look at decaf, see how we build the Z3 constraints. We want to have our own representation for constraints, instead of building them in Z3

PR #22, opened 2020-11-21
Begin interpreter design

mishazharov:
Addresses #7 

This PR creates headers for the interpreter

### DIFF
--- a/include/caffeine/Interpreter/Context.h
+++ b/include/caffeine/Interpreter/Context.h
@@ -1,0 +1,58 @@
+#ifndef CAFFEINE_INTERP_CONTEXT_H
+#define CAFFEINE_INTERP_CONTEXT_H
+
+#include <llvm/IR/Function.h>
+
+#include "caffeine/Interpreter/StackFrame.h"
+#include "caffeine/Solver/Solver.h"
+
+namespace caffeine {
+
+class Context {
+private:
+  /**
+   * This vector of assertions keeps the current set of invariants that we
+   * are keeping
+   */
+  std::vector<Assertion> assertions;
+
+public:
+  /**
+   * The context is constructed with the solver and the function that will be
+   * evaluated
+   */
+  Context(const Solver& solver, llvm::Function* func);
+
+  /**
+   * Create a new context that is independent from this
+   * one but has the same state.
+   */
+  Context fork() const;
+
+  /**
+   * Get the top frame of the stack.
+   *
+   * This should be used instead of directly manipulating the stack
+   * vector so that it continues to work when more advanced data
+   * structures are implemented.
+   */
+  StackFrame& stack_top();
+
+  /**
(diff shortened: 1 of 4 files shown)


include/caffeine/Interpreter/StackFrame.h
```diff
@@ -0,0 +1,57 @@
...
+  void insert(llvm::Value* value, ref<Operation>& expr);
+  /**
+   * Lookup a value within the current stack frame.
+   *
+   * There are two main cases here:
+   * 1. `value` is an existing variable
+   * 2. `value` is a constant
+   *
+   * In the first case we just look up the variable the `variables` map
+   * and then return it. In the second case we build a Z3 expression
```
Phantomical, line 46:
```suggestion
   * and then return it. In the second case we build an expression
```

include/caffeine/Interpreter/Context.h
```diff
@@ -0,0 +1,58 @@
...
+#include "caffeine/Solver/Solver.h"
+
+namespace caffeine {
+
+class Context {
+private:
+  /**
+   * This vector of assertions keeps the current set of invariants that we
+   * are keeping
+   */
```
Phantomical, line 16: Maybe just have this be something like
> The current set of invariants for this context.

include/caffeine/Interpreter/Context.h
```diff
@@ -0,0 +1,58 @@
...
+  std::vector<Assertion> assertions;
+
+public:
+  /**
+   * The context is constructed with the solver and the function that will be
+   * evaluated
+   */
+  Context(const Solver& solver, llvm::Function* func);
```
Phantomical, line 24: The context is going to have to own the solver so this will need to be either a `unique_ptr` or `shared_ptr`. I'd suggest a `shared_ptr` for the moment until we figure out more about how solvers are going to work.

include/caffeine/Interpreter/Context.h
```diff
@@ -0,0 +1,58 @@
...
+   *
+   * Does not modify the solver state. If this returns sat then you can
+   * get the solver model as a test case.
+   */
+  SolverResult check(const Assertion& expr);
+
+  /**
+   * Add a new assertion to the solver.
+   */
+  void add(const Assertion& assertion);
```
Phantomical, line 53: We probably want a different API for `add` and `check` since the new solver API is batch-oriented. Not exactly sure how it would look at the moment but it's something to think about.

You don't have to solve this one in this PR but I don't think decaf's API is gonna work well here.

Phantomical (approved):
Approved as long as you address the comments.

It's not something that needs to be fixed in this PR but I think you'll find that parts of decaf's context API aren't going to translate very well. When implementing the interpreter I think you'll want to do things directly with the solver for a while until you can figure out a good API.

PR #23, opened 2020-11-24
Add initial interpreter implementation

mishazharov:
This PR adds StackFrame and some other interpreter components

### DIFF
--- a/include/caffeine/Interpreter/Interpreter.h
+++ b/include/caffeine/Interpreter/Interpreter.h
@@ -32,11 +32,11 @@ public:
 
   ExecutionResult visitInstruction(llvm::Instruction &inst);
 
-  ExecutionResult visitPHINode(llvm::PHINode &node)       { CAFFEINE_ABORT(); };
-  ExecutionResult visitBranchInst(llvm::BranchInst &inst) { CAFFEINE_ABORT(); };
-  ExecutionResult visitReturnInst(llvm::ReturnInst &inst) { CAFFEINE_ABORT(); };
-  ExecutionResult visitCallInst(llvm::CallInst &inst)     { CAFFEINE_ABORT(); };
-  ExecutionResult visitSelectInst(llvm::SelectInst &inst) { CAFFEINE_ABORT(); };
+  ExecutionResult visitPHINode(llvm::PHINode &node)       { CAFFEINE_UNIMPLEMENTED(); };
+  ExecutionResult visitBranchInst(llvm::BranchInst &inst) { CAFFEINE_UNIMPLEMENTED(); };
+  ExecutionResult visitReturnInst(llvm::ReturnInst &inst) { CAFFEINE_UNIMPLEMENTED(); };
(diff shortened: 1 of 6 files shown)


src/Interpreter/Context.cpp
```diff
@@ -0,0 +1,13 @@
+#include "caffeine/Interpreter/Context.h"
```
mishazharov, line 1: feel free to ignore this file for now

src/Interpreter/Interpreter.cpp
```diff
@@ -0,0 +1,16 @@
+#include "caffeine/Interpreter/Interpreter.h"
+#include "caffeine/Support/Assert.h"
+
```
mishazharov, line 3: WIP

src/Interpreter/StackFrame.cpp
```diff
@@ -0,0 +1,26 @@
...
+  current_block = block;
+  current = block->begin();
+}
+
+void StackFrame::insert(llvm::Value *value, const ref<Operation> &expr) {
+  variables.insert_or_assign(value, expr);
+}
+
+ref<Operation> StackFrame::lookup(llvm::Value *value) const {
+  auto it = variables.find(value);
```
mishazharov, line 21: Removed the evaluate constant code since that will be taken care of by the solver

include/caffeine/Interpreter/Interpreter.h
```diff
@@ -1,29 +1,42 @@
...
   void execute();
 
-  ExecutionResult visitInstruction(ref<Operation>&);
+  ExecutionResult visitInstruction(llvm::Instruction &inst);
+
+  ExecutionResult visitPHINode(llvm::PHINode &node)       { CAFFEINE_ABORT(); };
+  ExecutionResult visitBranchInst(llvm::BranchInst &inst) { CAFFEINE_ABORT(); };
+  ExecutionResult visitReturnInst(llvm::ReturnInst &inst) { CAFFEINE_ABORT(); };
+  ExecutionResult visitCallInst(llvm::CallInst &inst)     { CAFFEINE_ABORT(); };
+  ExecutionResult visitSelectInst(llvm::SelectInst &inst) { CAFFEINE_ABORT(); };
```
Phantomical, line 39: nit: You might want to use `CAFFEINE_UNIMPLEMENTED` here as it has a better error message.

src/Interpreter/StackFrame.cpp
Phantomical: The problem here is that constants won't be found in the variables map. We still need the evaluate code for that.

I agree that constant folding can probably be done by the solver for now (for performance reasons we may want to change this in the future but I don't think it's worth it at this point).

Phantomical (commented):
Just a few small suggestions. Will review again once the PR is complete :)

Phantomical (approved):
This LGTM. We'll probably have to change up things as we go through with the implementation but that's fine.

PR #25, opened 2020-11-29
Implement a check-format target

Phantomical:
Our current check-format pass in CI doesn't work. This PR adds a target that will accurately check whether all files in the project are formatted and, in the case where it finds an unformatted file, print the diff between the formatted and unformatted files as the error message.

Note that this will only work on systems that have diff available on the command line.


mishazharov (approved):
LGTM

PR #30, opened 2020-11-30
Add Phi node and branch implementation

mishazharov:
This PR implements PHI nodes and branching in a similar way to decaf

### DIFF
--- a/src/Interpreter/Interpreter.cpp
+++ b/src/Interpreter/Interpreter.cpp
@@ -46,4 +46,58 @@ ExecutionResult Interpreter::visitAdd(llvm::BinaryOperator& op) {
   return ExecutionResult::Continue;
 }
 
+ExecutionResult Interpreter::visitPHINode(llvm::PHINode& node) {
+  auto &frame = ctx->stack_top();
+
+  // PHI nodes in the entry block is invalid.
+  CAFFEINE_ASSERT(frame.prev_block != nullptr);
+
+  auto value = frame.lookup(node.getIncomingValueForBlock(frame.prev_block));
+  frame.insert(&node, value);
+
+  return ExecutionResult::Continue;
+}
+
+ExecutionResult Interpreter::visitBranchInst(llvm::BranchInst &inst) {
+  if (!inst.isConditional()) {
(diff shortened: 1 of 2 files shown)


src/Interpreter/Interpreter.cpp
```diff
@@ -45,4 +45,58 @@ ExecutionResult Interpreter::visitAdd(llvm::BinaryOperator& op) {
...
+  if (!inst.isConditional()) {
+    ctx->stack_top().jump_to(inst.getSuccessor(0));
+    return ExecutionResult::Continue;
+  }
+
+  auto& frame = ctx->stack_top();
+  auto cond = frame.lookup(inst.getCondition());
+
+  auto is_t = ctx->check(Assertion::constant(cond != 0));
+  auto is_f = ctx->check(Assertion::constant(cond == 0));
```
mishazharov, line 70: Not sure if this is the right way to check our conditions since the `!=` and `==` overloads seem to only work for constants

Phantomical: Yeah this isn't what you want.

It should be something like 
```
auto is_t = ctx->check(Assertion(cond));
auto is_f = ctx->check(!Assertion(cond));
```

src/Interpreter/Interpreter.cpp
```diff
@@ -45,4 +45,58 @@ ExecutionResult Interpreter::visitAdd(llvm::BinaryOperator& op) {
...
+  //       equivalent to sat. Maybe future branches will bring the
+  //       equation back to being solvable.
+  if (is_t != SolverResult::UNSAT && is_f != SolverResult::UNSAT) {
+    auto fork = ctx->fork();
+
+    // In cases where both conditions are possible we follow the
+    // false path. This should be enough to get us out of most loops
+    // and actually exploring the rest of the program.
+    fork.add(cond);
+    ctx->add(Assertion::constant(!cond));
```
Phantomical, line 82: Same thing here, this doesn't do what you want.

It should be `ctx->add(!Assertion(cond))`.

Same with all of the others.

Phantomical (approved):
You've got a few mistakes with the assertions. Approved conditional on those being fixed. LGTM otherwise

PR #32, opened 2020-12-01
[Memory 1/N] Create slot_map data structure

Phantomical:
See the doc comment for slotmap for an explanation of what this does and the comment within for an explanation of how it's implemented. I've included some tests that should cover most of the general use, although they might not cover all of the edge cases.

This is needed as part of the memory system.

### DIFF
--- a/include/caffeine/ADT/SlotMap.h
+++ b/include/caffeine/ADT/SlotMap.h
@@ -11,6 +11,12 @@
 namespace caffeine {
 
 namespace detail {
+  /**
+   * RAII class that calls a user-provided function during it's destructor.
+   *
+   * This is useful for ensuring that code is executed even if an exception is
+   * thrown (i.e. it can be used to emulate a try/finally statenent).
+   */
   template <typename F>
   class drop_guard : F {
   public:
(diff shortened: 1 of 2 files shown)


include/caffeine/ADT/SlotMap.h
```diff
@@ -0,0 +1,449 @@
...
+#include <deque>
+#include <optional>
+#include <type_traits>
+#include <utility>
+
+namespace caffeine {
+
+namespace detail {
+  template <typename F>
+  class drop_guard : F {
```
mishazharov, line 15: What's this class for? Could use more documentation

include/caffeine/ADT/SlotMap.h
```diff
@@ -0,0 +1,449 @@
...
+   */
+  template <typename T>
+  constexpr T replace(T& loc, const T& value) {
+    T old = std::move(loc);
+    loc = value;
+    return old;
+  }
+
+  template <typename T>
+  using remove_cvref_t = std::remove_cv_t<std::remove_reference_t<T>>;
```
mishazharov, line 45: What's this responsible for? Can't find any other references to it

mishazharov (approved):
LGTM otherwise

include/caffeine/ADT/SlotMap.h
Phantomical: This was used at some point during development but I guess that code got rewritten so it's just dead code then.

PR #35, opened 2020-12-01
[Memory 4/N] Implement allocation-related operations.

Phantomical:
This introduces 3 new derived opcode classes:
- `AllocOp` - allocates a new array that is initialized with copies of the same byte (meant to be used by alloca/malloc)
- `LoadOp` - loads a byte from an existing array
- `StoreOp` - returns an array with the byte at a given position overwritten with a new one

Beyond that, it adds support for numbered constants. Certain constants may not be user-facing (e.g. memory allocation addresses) and so they may not have a name that can easily be assigned. This mirrors Z3's API where constants can be either named or numbered. Also included is support for these new opcodes in `OpVisitor`

These are enough to implement the rest of the memory model. As part of this PR I also ended up compacting the opcode space somewhat which sorta blew up the number of included lines.

This PR depends on #34 so it currently includes the changes of that PR.

### DIFF
--- a/include/caffeine/IR/Operation.h
+++ b/include/caffeine/IR/Operation.h
@@ -95,6 +95,12 @@ enum class FCmpOpcode : uint8_t {
  *    Value type.
  */
 class Operation {
+protected:
+  // Base opcode used for FCmp opcodes
+  static constexpr uint16_t fcmp_base = 4;
+  // Base opcode used for ICmp opcodes
+  static constexpr uint16_t icmp_base = 5;
+
 public:
   /**
    * Notes on bit representation:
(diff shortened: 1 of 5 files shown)


include/caffeine/IR/Operation.h
```diff
@@ -545,6 +589,80 @@ class FCmpOp : public BinaryOp {
...
 };
 
+/**
+ * Memory allocation operation.
+ *
+ * This operation creates a new symbolic array that is filled with a default
+ * value (of type i8). It is meant to be used for modelling allocations (e.g.
+ * malloc or alloca).
+ */
+class AllocOp : public Operation {
```
mishazharov, line 599: Do we need a deallocation op?

mishazharov (approved):

include/caffeine/IR/Operation.h
Phantomical: Nope, deallocation is handled at a level above this one.

PR #36, opened 2020-12-02
Make everything compile and add main executable

Phantomical:
This PR has the following changes
- Add all the methods that were used be we forgot to implement
- Add a interface for logging failures
- Implement that interface to print out the constants in the model when executed
- More or less copy over the main binary

I have no idea if this runs because we haven't implemented enough of an interpreter to actually test anything.


mishazharov (approved):
:ship:

PR #37, opened 2020-12-03
Implement more operations for the interpreter

Phantomical:
This commit implements most of the non-controlflow opcodes we'll need for the demo although there's still a number left.

If we implement call and the builtins then we should be able to reuse the test suite from last time.


mishazharov (approved):
Can't review rn but can slap approve

Phantomical:
That's good enough! It's mostly copy-paste code anyway

PR #44, opened 2020-12-04
Add warnings if we don't find a clang installation

Phantomical:
As in title.


mishazharov (approved):

test/CMakeLists.txt
```diff
@@ -2,6 +2,13 @@
 find_program(CLANG clang)
 find_program(CLANGXX clang++)
```
mishazharov, line 3: honestly we should probably make this required

Phantomical: For now I'll settle with a warning and we can revisit it later.

Theoretically it's not necessary to build caffeine so I don't think it really needs to be required.

PR #45, opened 2020-12-04
Add more tests

mishazharov:
Need to implement select instruction first


mishazharov:
This basically got rolled into other PRs

PR #47, opened 2020-12-04
Add first pass of Z3Solver SelectOp

mishazharov:
This is necessary to implement the SelectOp in the interpreter, and pass the collatz test case

### DIFF
--- a/src/Solver/Z3Solver.cpp
+++ b/src/Solver/Z3Solver.cpp
@@ -38,6 +38,15 @@ static z3::expr normalize_to_bool(const z3::expr& expr) {
   return expr;
 }
 
+static z3::expr normalize_to_int(const z3::expr& expr) {
+  auto sort = expr.get_sort();
+
+  if (sort.is_bv() && sort.bv_size() == 1)
+    return expr == 1;
+
+  return expr;
+}
+
 class EmptyModel : public Model {
 public:
   EmptyModel(SolverResult result) : Model(result) {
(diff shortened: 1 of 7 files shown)


test/run-fail/sdiv-udiv-inequal.c
```diff
@@ -0,0 +1,26 @@
+
+#include "caffeine.h"
+#include <stdint.h>
+
+uint32_t sdiv(uint32_t x, uint32_t y) {
+  // Don't have or yet. Need to emulate it.
+  if (x == INT32_MIN)
+    caffeine_assume(y != -1);
```
Phantomical, line 8: We have `or` now. In the interest of clarity why don't we change this to
```
caffeine_assume(x != INT32_MIN || y != -1);
```

test/run-fail/sdiv-zero.c
```diff
@@ -0,0 +1,12 @@
+
+#include "caffeine.h"
+#include <stdint.h>
+
+// Test that signed division checks for divide-by-zero.
+int32_t test(int32_t x, int32_t y) {
+  // Don't have or yet. Need to emulate it.
+  if (x == INT32_MIN)
```
Phantomical, line 8: Same comment about updating the assume statement here.

src/Solver/Z3Solver.cpp
```diff
@@ -38,6 +38,15 @@ static z3::expr normalize_to_bool(const z3::expr& expr) {
   return expr;
 }
 
+static z3::expr normalize_to_int(const z3::expr& expr) {
```
Phantomical, line 41: We're going to end up with a couple of these up here as parallel changes get merged :)

Not a problem for this PR. Whomever merges last will have to convert.

Phantomical (approved):
LGTM! There's 2 small cleanups you might want to do for the test cases since we support more operations then when they were written for decaf.

mishazharov:
I'll have to add these after adding the interpreter SelectOp as well

PR #48, opened 2020-12-04
Start getting memory code tested

Phantomical:
This is the set of changes needed to get memory allocation code working. I'm not sure if it's completely correct at this point but it's more correct than it was when I started this.

As a bonus this makes `run-fail/sdiv-udiv-inequal.c` about 20 times faster!

### DIFF
--- a/src/Solver/Z3Solver.h
+++ b/src/Solver/Z3Solver.h
@@ -1,0 +1,89 @@
+
+#pragma once
+
+#include "caffeine/Solver/Z3Solver.h"
+
+#include "caffeine/IR/Assertion.h"
+#include "caffeine/IR/Operation.h"
+#include "caffeine/IR/Value.h"
+#include "caffeine/IR/Visitor.h"
+
+#include <z3++.h>
+
+#include <string>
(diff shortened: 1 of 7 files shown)


src/Solver/Z3Solver.h
```diff
@@ -0,0 +1,89 @@
+
+#pragma once
```
mishazharov, line 2: Why did this get moved into `src`?

Phantomical: Two main reasons
1. None of `Z3Model` or `Z3OpVisitor` are actually useful to downstream consumers so it's easier if they're not publically exposed.
2. `z3++.h` manages to include just about every single slow-to-compile header from the C++ standard library so it adds 5-10s of compilation time to any cpp file that it ends up being included in. In the interests of keeping compile times down as we have more and more source files I would rather have it in a private file that's only included in a few places.

Basically, the parts I moved in to `src` are internal implementation details and slow down compilation so they're better off in a private header file.

mishazharov (approved):

PR #54, opened 2020-12-04
Fix typo within const slot_map::end

Phantomical:
This meant that any time we tried to iterate over a const `slot_map `it wouldn't return any values since we would always have that `std::begin(map) == std::end(map)`.

I've added an iteration test case although it wouldn't have caught this typo (it tests non-const iteration, the bug was in the const versions of the methods)


mishazharov (approved):
Maybe you should add a const test case as well? Your discretion

PR #58, opened 2020-12-04
Add cmake code to allow skipping tests

Phantomical:
`select` isn't implemented so collatz cannot run. This adds cmake code to allow us to skip tests by putting `SKIP TEST` somewhere on the first line and then uses that to skip `collatz.c`.

### DIFF
--- a/test/run-fail/collatz.c
+++ b/test/run-fail/collatz.c
@@ -25,8 +25,5 @@ uint32_t __attribute__((noinline)) collatz(uint32_t x) {
 }
 
 void test(uint32_t n) {
-  // Select not implemented yet; skip
-  return;
-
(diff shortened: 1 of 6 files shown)


mishazharov:
I just added select in another PR, if we merge that we can close this

test/run-fail/collatz.c
```diff
@@ -1,3 +1,4 @@
+// SKIP TEST
```
mishazharov, line 1: If we remove this line it should work

mishazharov (approved):

Phantomical:
Not closing as I think the test skipping stuff is worth having. I have removed the SKIP TEST from collatz though.

PR #61, opened 2020-12-05
Add bitwise operations

mishazharov:


Phantomical (approved):
LGTM! You might want to implement them for the default `Model::evaluate` method in `Solver.cpp` but that's not essential for the demo so it can probably be done later.

mishazharov:
Will create a ticket for it

PR #64, opened 2020-12-05
Give failure logger required context to see all assertions

Phantomical:
The failure logger wasn't able to see the assertion used as an argument to `check`. This lead to not printing values that should have been asserted when running the fnv.c test case.

I also went through and cleaned up a few test cases that were coded to restrictions in decaf that no longer apply.


mishazharov (approved):
Nice

PR #69, opened 2020-12-06
(Mostly) make floats work

Phantomical:
This makes floating point values and computations work in most cases. Here's the list of what was changed
- Implement support for LLVM `fcmp` instruction
- Use Z3's builtin `isnan` checking functions since `x == x` apparently doesn't work as expected with NaNs.
- Add support for extracting floating point values to `llvm::APFloat`
- Allow creating constants from LLVM floating point literals.
- Add a test case verifying that `x != x` fails with NaNs.

There's still some quirks of LLVM that are unsupported here but those can go in a separate PR. (I do have uncommitted test cases which fail there).

### DIFF
--- a/src/Interpreter/StackFrame.cpp
+++ b/src/Interpreter/StackFrame.cpp
@@ -14,6 +14,12 @@ ref<Operation> evaluate_constant(const llvm::Constant* constant) {
     return ConstantInt::Create(value);
   }
 
+  if (auto* fpconst = llvm::dyn_cast<llvm::ConstantFP>(constant)) {
+    const llvm::APFloat& value = fpconst->getValueAPF();
+
+    return ConstantFloat::Create(value);
+  }
+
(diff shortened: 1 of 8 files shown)


src/Interpreter/StackFrame.cpp
```diff
@@ -14,6 +14,12 @@ ref<Operation> evaluate_constant(const llvm::Constant* constant) {
     return ConstantInt::Create(value);
   }
 
+  if (auto* fpconst = llvm::dyn_cast<llvm::ConstantFP>(constant)) {
+    const llvm::APFloat& value = fpconst->getValueAPF();
+
+    return ConstantFloat::Create(value);
+  }
+
   // We only implement integers at the moment
```
mishazharov, line 23: comment is out of date :-)

mishazharov (approved):

PR #70, opened 2021-01-01
Add support for floating-point opcodes

Phantomical:
This adds support for floating point arithmetic opcodes and also uses the Z3 C interface for floating point operations that are not supported by the C++ interface.

Note: Conversion from Z3 values to APFloat is currently broken when the Z3 fpa is an infinity. Will hold off merging until I fix that.

### DIFF
--- a/src/Interpreter/Interpreter.cpp
+++ b/src/Interpreter/Interpreter.cpp
@@ -217,6 +217,47 @@ ExecutionResult Interpreter::visitNot(llvm::BinaryOperator& op) {
   return ExecutionResult::Continue;
 }
 
+ExecutionResult Interpreter::visitFAdd(llvm::BinaryOperator& op) {
+  StackFrame& frame = ctx->stack_top();
+
+  auto lhs = frame.lookup(op.getOperand(0));
+  auto rhs = frame.lookup(op.getOperand(1));
+
+  frame.insert(&op, BinaryOp::CreateFAdd(lhs, rhs));
+
+  return ExecutionResult::Continue;
+}
+ExecutionResult Interpreter::visitFSub(llvm::BinaryOperator& op) {
+  StackFrame& frame = ctx->stack_top();
+
+  auto lhs = frame.lookup(op.getOperand(0));
+  auto rhs = frame.lookup(op.getOperand(1));
+
+  frame.insert(&op, BinaryOp::CreateFAdd(lhs, rhs));
+
+  return ExecutionResult::Continue;
+}
+ExecutionResult Interpreter::visitFMul(llvm::BinaryOperator& op) {
+  StackFrame& frame = ctx->stack_top();
+
+  auto lhs = frame.lookup(op.getOperand(0));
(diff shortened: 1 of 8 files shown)


src/Interpreter/Interpreter.cpp
```diff
@@ -217,6 +217,47 @@ ExecutionResult Interpreter::visitNot(llvm::BinaryOperator& op) {
...
+
+  return ExecutionResult::Continue;
+}
+ExecutionResult Interpreter::visitFDiv(llvm::BinaryOperator& op) {
+  StackFrame& frame = ctx->stack_top();
+
+  auto lhs = frame.lookup(op.getOperand(0));
+  auto rhs = frame.lookup(op.getOperand(1));
+
+  frame.insert(&op, BinaryOp::CreateFAdd(lhs, rhs));
```
mishazharov, line 256: Do we need to check for division by 0?

src/Solver/Z3Solver.cpp
```diff
@@ -277,7 +277,12 @@ z3::expr Z3OpVisitor::visitConstantInt(const ConstantInt& op) {
 
 z3::expr Z3OpVisitor::visitConstantFloat(const ConstantFloat& op) {
   // TODO: Reimplement this correctly
-  return ctx->fpa_val(op.value().convertToFloat());
+  auto expr = z3::expr(
```
mishazharov, line 280: Is the TODO still accurate?

mishazharov (approved):
Otherwise LGTM

src/Solver/Z3Solver.cpp
```diff
@@ -378,6 +383,27 @@ static z3::expr is_nan(const z3::expr& e) {
...
+  val.check_error();
+  return val;
+}
+static z3::expr fpa_geq(const z3::expr& a, const z3::expr& b) {
+  auto val = z3::expr(a.ctx(), Z3_mk_fpa_geq(a.ctx(), a, b));
+  val.check_error();
+  return val;
+}
+static z3::expr fpa_gt(const z3::expr& a, const z3::expr& b) {
+  auto val = z3::expr(a.ctx(), Z3_mk_fpa_leq(a.ctx(), a, b));
```
mishazharov, line 402: `Z3_mk_fpa_leq` => `Z3_mk_fpa_gt`?

src/Interpreter/Interpreter.cpp
Phantomical: Nope. FP division by zero just returns infinity so it's all good.

src/Solver/Z3Solver.cpp
Phantomical: Yup. The above just converts to a double. Any bigger FP types will get silently truncated when they go through this.

src/Solver/Z3Solver.cpp
Phantomical: Nice catch! Fixed.

PR #71, opened 2021-01-01
Make Type printable

Phantomical:
I found that I wanted this while doing some debugging. Doesn't work for all types just yet but it does work for the ones that we care about most of the time.

### DIFF
--- a/src/IR/Type.cpp
+++ b/src/IR/Type.cpp
@@ -156,6 +156,8 @@ std::ostream& operator<<(std::ostream& os, const Type& t) {
     return os << "i" << t.bitwidth();
 
   if (t.is_float())
+    // TODO: Need a way to distinguish between same-size floats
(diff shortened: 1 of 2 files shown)


src/IR/Type.cpp
```diff
@@ -146,4 +148,23 @@ Type Type::from_llvm(llvm::Type* type) {
...
+  if (t.is_float())
+    return os << "f" << (t.mantissa_bits() + t.exponent_bits());
+
+  if (t.is_pointer())
+    return os << "void*";
+
+  if (t.is_array())
+    return os << "array";
+
+  CAFFEINE_UNIMPLEMENTED();
```
mishazharov, line 167: Maybe instead of throwing we should print a `???` or something like that since this isn't really an unrecoverable situation

mishazharov (approved):

PR #73, opened 2021-01-02
Implement undef operation

Phantomical:
I've been trying to implement vector operations and found that I needed this. I'm pulling it out here so that it can be reviewed separately.

For the moment I've approximated the semantics of the undef opcode within Z3 by just returning a constant 0. In the future we should probably look deeper into what's needed to properly handle undefs in LLVM.


mishazharov (approved):
LGTM!

PR #74, opened 2021-01-03
Implement basic constant folding

Phantomical:
This implements constant folding as part of expression construction. That is, when you call the corresponding `CreateXXX` method with constant operands it'll automatically do the proper constant folding. This means that no other code needs to be changed to take advantage of this.

This PR is dependant on #73 and includes its commits as well.


mishazharov (approved):
lgtm as well

PR #80, opened 2021-01-03
Simplify Operation internals

Phantomical:
I got fed up with trying to figure out all the needed moving parts to add new internal state to operation and decided to simplify the whole thing somewhat.


mishazharov (approved):
Nice refactor

PR #81, opened 2021-01-06
Add ContextValue

Phantomical:
This is an attempt to break apart bits of #76 into more reviewable chunks.

This part adds a new value representation for program values (which can be vectors or scalars at the moment). It isn't currently used anywhere within the program but will get used in the next PR. It also adds a method to map over these new values which is meant to make it easier to implement the visitor method. For examples of how it's used see the other PR.


mishazharov (approved):
Makes sense, lgtm

PR #84, opened 2021-01-06
Add ConstantArray operation

Phantomical:
This is a prerequisite for supporting just about any other memory operations. I've also modified the array type so that it stores the bitwidth of the integer type used to index into the array.

This is a first step towards closing #82.

### DIFF
--- a/include/caffeine/IR/Operation.inl
+++ b/include/caffeine/IR/Operation.inl
@@ -27,6 +27,7 @@ namespace caffeine {
 // All derived operation types should be the same size
 static_assert(sizeof(ConstantInt) == sizeof(Operation));
 static_assert(sizeof(ConstantFloat) == sizeof(Operation));
+static_assert(sizeof(ConstantArray) == sizeof(Operation));
 static_assert(sizeof(Constant) == sizeof(Operation));
 static_assert(sizeof(BinaryOp) == sizeof(Operation));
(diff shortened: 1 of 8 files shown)


include/caffeine/IR/Operation.inl
```diff
@@ -259,6 +260,14 @@ inline const llvm::APFloat& ConstantFloat::value() const {
   return std::get<llvm::APFloat>(inner_);
 }
 
+/***************************************************
+ * ConstantArray                                   *
+ ***************************************************/
+inline llvm::ArrayRef<char> ConstantArray::data() const {
+  const std::string& str = std::get<std::string>(inner_);
```
mishazharov, line 267: is string a good representation for arbitrary data? I'm wondering if we can have encoding issues etc.

mishazharov (approved):

include/caffeine/IR/Operation.inl
Phantomical: The only difference between a `std::string` and a `std::vector<char>` is that the string will be nul-terminated. As long as we don't use `c_str()` nothing in `std::string` breaks if a nul byte shows up (and we can enforce this by not using `c_str` when we know that the string might contain arbitrary data). Other than that `std::string` doesn't do anything encoding-wise so I think we're good?

Issue #90, opened 2021-01-06
Investigate simplifying the FCmp operation


Phantomical:
Right now our `FCmp` operation mirrors the semantics of LLVM's `fcmp` operation. This is somewhat less than ideal since it results in more complicated expressions in the underlying solver without giving constant-folding a chance to simplify it.

We may be better off if we split the current `FCmp` into two distinct operations:
- `FCmp`: Has EQ, NE, LT, LE, GT, and GE where behaviour for NaNs is completely unspecified. (i.e. can be whatever, doesn't even have to be consistent between instances)
- `IsNan`: Unary operation that just checks whether the operand passed in is NaN.

Then we can implement the bulk of the logic in the interpreter (which allows constant folding to simplify it). This should also simplify the implementations of solvers.

PR #95, opened 2021-01-10
Implement FIsNaN opcode

mishazharov:
This is the first part of #90 in which we split `FCmp` into two distinct operations


Phantomical (approved):
LGTM! 

Once you've finished with this issue you might also want to look at implementing #96 :)

PR #101, opened 2021-01-12
Implement getelementptr instruction

### LINKED ISSUE #100: Implement getelementptr instruction

Phantomical:
As in description. The getelementptr instruction corresponds to member accesses, array accesses, etc. At the moment I haven't implemented support for vector getelementptr instructions since I'm not entirely sure how they work.

This PR depends on #99, #98, and #97 and includes all of their changes. I'll rebase to remove those as they get approved.

Closes #100.

### DIFF
--- a/src/Interpreter/Interpreter.cpp
+++ b/src/Interpreter/Interpreter.cpp
@@ -7,6 +7,7 @@
 #include <boost/range/combine.hpp>
 #include <boost/range/iterator_range.hpp>
 #include <fmt/format.h>
+#include <llvm/IR/GetElementPtrTypeIterator.h>
 #include <llvm/Support/raw_ostream.h>
 
 #include <iostream>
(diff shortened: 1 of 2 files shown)


src/Interpreter/Interpreter.cpp
```diff
@@ -640,6 +641,70 @@ Interpreter::visitShuffleVectorInst(llvm::ShuffleVectorInst& inst) {
...
 }
 
+static llvm::Type* vector_inner_type(llvm::Type* type) {
+  while (type->isVectorTy())
+    type = type->getVectorElementType();
+  return type;
+}
+
+ExecutionResult
+Interpreter::visitGetElementPtrInst(llvm::GetElementPtrInst& inst) {
```
mishazharov, line 651: It would be good to add a simple test here. Maybe the llvm IR straight from https://llvm.org/docs/LangRef.html#getelementptr-instruction with a comparison at the end?

mishazharov (approved):

src/Interpreter/Interpreter.cpp
Phantomical: I would but I'm actually unable to create a test case that uses `getelementptr` and also doesn't use something we don't support.

I'll add some test cases once I add support for the needed instructions.

PR #102, opened 2021-01-12
Implement support for load instructions

### LINKED ISSUE #51: Implement load instruction
These should basically consist of
1. Check to see if the pointer could possibly be outside of any allocation (via `MemHeap::check_valid`). If so, abort the context with an error.
2. Resolve the pointer to a set of allocations, fork for each resolved allocation.
3. Verify that the offset is in-bounds for the allocation.
4. Call `Allocation::read`/`Allocation::write` to get the value read from the array.

Phantomical:
This PR implements the final few bits for supporting the two test cases I've added here. Namely, it
- Implements the load instruction
- (I don't entirely remember the rest - will update once I've rebased and other PRs have been merged)

This currently depends on #97, #98, #99, and #101.

Closes #51 and addresses half of #52.


src/Solver/Z3Solver.h
```diff
@@ -93,11 +101,27 @@ class Z3OpVisitor : public ConstOpVisitor<Z3OpVisitor, z3::expr> {
...
   z3::expr visitNot (const UnaryOp& op);
   z3::expr visitFNeg(const UnaryOp& op);
   z3::expr visitFIsNaN(const UnaryOp& op);
   // clang-format on
+
+  /**
+   * When a temporary constant is needed then use this function.
+   *
+   * It explicitly creates constants with high-integer names that won't collide
+   * unless the running program creates more than 2^29 constant names.
```
mishazharov, line 116: nit: Sounds like a hack, is it possible to create constants with some kind of prefix? Doesn't really matter but if it's an easy fix then might be worth doing that

src/Interpreter/Interpreter.cpp
```diff
@@ -704,6 +704,56 @@ Interpreter::visitGetElementPtrInst(llvm::GetElementPtrInst& inst) {
...
+    auto value = alloc.read(resolved[i].offset(), load_ty, layout);
+    forked.stack_top().insert(&inst, value);
+  }
+
+  Allocation& alloc = ctx->heap()[resolved[0].alloc()];
+  ctx->add(
+      alloc.check_inbounds(resolved[0].offset(), load_ty.byte_size(layout)));
+
+  auto value = alloc.read(resolved[0].offset(), load_ty, layout);
+  ctx->stack_top().insert(&inst, value);
```
mishazharov, line 753: Might be worth breaking this down into a macro?

mishazharov (approved):
Just a few nits, otherwise LGTM

src/Solver/Z3Solver.h
Phantomical: I'd rather not have to allocate a string every time we want to create a temporary constant. Instead I've put an assertion where constant names are created within `Context`. If that ever fires then we can revisit this with a better solution?

I think 2^29 temporary constant numbers is probably never going to be hit. However, if it does become a problem then we can solve it then?

src/Interpreter/Interpreter.cpp
Phantomical: Good idea. Done.

src/Solver/Z3Solver.h
mishazharov: Yeah, sounds good to me!

PR #109, opened 2021-01-13
Simplify FCmp part two

### LINKED ISSUE #90: Investigate simplifying the FCmp operation

mishazharov:
This PR adds the second part of moving FCmp logic into the interpreter

Closes #90


src/IR/Operation.cpp
```diff
@@ -700,8 +706,8 @@ ref<Operation> UnaryOp::CreateNot(const ref<Operation>& operand) {
   return Create(Opcode::Not, operand);
 }
 
-DECL_UNOP_CREATE(FNeg, ASSERT_FP);
-DECL_UNOP_CREATE(FIsNaN, ASSERT_FP);
+DECL_UNOP_CREATE(FNeg, ASSERT_FP, Type::float_ty(11, 53));
```
Phantomical, line 709: This doesn't work if the input isn't a double.

src/IR/Operation.cpp
```diff
@@ -801,8 +807,12 @@ ref<Operation> SelectOp::Create(const ref<Operation>& cond,
   CAFFEINE_ASSERT(true_value, "true_value was null");
   CAFFEINE_ASSERT(false_value, "false_value was null");
 
-  CAFFEINE_ASSERT(cond->type() == Type::int_ty(1),
-                  "select condition was not an i1");
+  if (cond->type() != Type::int_ty(1)) {
+    std::stringstream s;
+    s << "select condition was not an i1, it was " << cond->type();
+    CAFFEINE_ASSERT(cond->type() == Type::int_ty(1), s.str());
+  }
```
Phantomical, line 814: Nit: If you use `fmt::format` you can put this all in the message position in the macro and avoid having to write out the if statement.

src/Interpreter/Interpreter.cpp
```diff
@@ -368,7 +379,7 @@ ExecutionResult Interpreter::visitFCmpInst(llvm::FCmpInst& fcmp) {
     FCMP_CASE(UGE);
     FCMP_CASE(ULT);
     FCMP_CASE(ULE);
-    FCMP_CASE(UNE);
+    FCMP_CASE_DEF(UNE, true);
```
Phantomical, line 382: Don't you need to do this for all the possible comparisons?

src/Solver/Z3Solver.cpp
```diff
@@ -449,46 +449,47 @@ z3::expr Z3OpVisitor::visitFCmp(const FCmpOp& op) {
...
-    expr = fpa_leq(lhs, rhs) || is_nan(lhs) || is_nan(rhs);
+    expr = fpa_leq(lhs, rhs);
     break;
   case FCmpOpcode::UNE:
-    expr = lhs != rhs || is_nan(lhs) || is_nan(rhs);
+    expr = lhs != rhs;
     break;
   case FCmpOpcode::UNO:
-    expr = is_nan(lhs) || is_nan(rhs);
+    expr = ctx->bv_val(0, 1);
```
Phantomical, line 492: I think you can just rip out `FCmpOpcode::UNO/ORD` since they don't apply anymore.

Phantomical (changes requested):
Looks mostly good! There's two bigger issues and I've left a few nits.

You might also want to add a test case that tests some of the ordered fcmp variants.

Fix the main issues and I'll approve. Nits are at your discretion.

src/Interpreter/Interpreter.cpp
```diff
@@ -346,14 +346,25 @@ ExecutionResult Interpreter::visitFCmpInst(llvm::FCmpInst& fcmp) {
...
+                       ref<Operation> neitherIsNaN = SelectOp::Create(         \
+                           UnaryOp::CreateFIsNaN(lhs), def,                    \
+                           SelectOp::Create(UnaryOp::CreateFIsNaN(rhs), def,   \
+                                            thenFcmp));                        \
+                       return neitherIsNaN;                                    \
+                     },                                                        \
+                     lhs, rhs));                                               \
+    return ExecutionResult::Continue;
+
+#define FCMP_CASE(op) FCMP_CASE_DEF(op, false)
```
Phantomical, line 367: Nit: I'd get rid of this and always use FCMP_CASE_DEF since it's more explicit about whats going on.

Phantomical (approved):
LGTM! Can you also remove `UNO` and `ORD` from `FCmpOpcode` before merging?

PR #115, opened 2021-01-16
Implement inttoptr and ptrtoint

Phantomical:
Mostly as in title. Here's the detailed list of changes:
- Added some utility operation factory functions (`CreateTruncOrZExt` and `CreateTruncOrSExt`) that behave as named
- Implement inttoptr
- Implement ptrtoint
- Fix the constant evaluation logic for both opcodes

### DIFF
--- a/src/Interpreter/CtxConstEval.cpp
+++ b/src/Interpreter/CtxConstEval.cpp
@@ -110,20 +110,27 @@ static ContextValue evaluate_expr(Context* ctx, llvm::ConstantExpr* expr) {
   case Instruction::FPToSI:   return CAST_OP(UnaryOp::CreateFpToSI(type, value));
     // clang-format on
 
-  case Instruction::IntToPtr:
+  case Instruction::IntToPtr: {
+    auto layout = ctx->llvm_module()->getDataLayout();
     return transform_value(
         [=](const auto& value) {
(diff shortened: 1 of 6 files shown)


src/Interpreter/CtxConstEval.cpp
```diff
@@ -110,20 +110,28 @@ static ContextValue evaluate_expr(Context* ctx, llvm::ConstantExpr* expr) {
...
-  case Instruction::IntToPtr:
+  case Instruction::IntToPtr: {
+    auto layout = ctx->llvm_module()->getDataLayout();
     return transform_value(
         [=](const auto& value) {
+          return ContextValue(Pointer(UnaryOp::CreateTruncOrZExt(
+              Type::int_ty(layout.getPointerSizeInBits(
+                  expr->getType()->getPointerAddressSpace())),
+              value.scalar())));
           return ContextValue(Pointer(value.scalar()));
```
mishazharov, line 121: If I'm not mistaken you forgot to delete a line?

mishazharov (approved):

src/Interpreter/CtxConstEval.cpp
Phantomical: Good catch! Fixed.

PR #117, opened 2021-01-16
Rewrite cmake build steps for LLVM IR makefiles

Phantomical:
When writing the code for malloc and free I needed to be able to link some runtime IR into the final bitcode for test cases. This PR adds the required code to be able to have LLVM IR libraries that are linked together.

It works by defining new cmake languages that compile to llvm bitcode and a linker language that shells out to llvm-link to link it all together. I've also rewritten the test cmake files to use the new methods.

See the explanatory comments within `LLVMIRUtils.cmake` for a bit of how it works but TBH it's probably not worth trying to understand it all.


mishazharov (approved):
> but TBH it's probably not worth trying to understand it all.

😎 ship it

PR #113, opened 2021-01-14
Implement store instruction

Phantomical:
As in title. Also had to fix `MemHeap::write` but everything seems to work now.

Closes #108.

PR #119, opened 2021-01-16
Implement bitcast instruction

Phantomical:
This implements a limited version of the bitcast instruction that only supports pointers. 

I've added a test case but it's currently marked as skipped because #113 hasn't landed.

### DIFF
--- a/test/run-pass/mem/pointer-bitcast.c
+++ b/test/run-pass/mem/pointer-bitcast.c
@@ -1,0 +1,11 @@
+
+#include "caffeine.h"
+#include <stdint.h>
+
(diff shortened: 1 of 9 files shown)


test/run-pass/mem/pointer-bitcast.c
```diff
@@ -0,0 +1,12 @@
+// SKIP TEST
```
mishazharov, line 1: #113 is approved so let's unskip this

mishazharov (approved):

Phantomical:
I tried running this now and it's broken so I'm going to try and fix it before merging

PR #131, opened 2021-01-20
Implement alloca instruction

### LINKED ISSUE #50: Implement Alloca instruction
This creates a temporary allocation for the lifetime of the stack frame. We'll need to modify the current stack frame to track these temporary allocations as well.

Phantomical:
Here's what was changed:
- We now have support for the `llvm.lifetime.(start|end).*` intrinsics. They have no effects and so they're ignored.
- Allocations now have a kind to verify that the right deallocation method is being used. (e.g. so we can emit errors if someone tries to free a global or stack allocation)
- Extra logic to delete stack allocations upon returning from a function
- And of course, implementing the alloca instruction itself

I've also discovered that cmake versions 3.17+ allow you to add flags to the ctest invocation so I've configured it to add the `--output-on-default` flag if using a supported version.

Closes #50.

### DIFF
--- a/test/run-fail/mem/free-stack-alloca.c
+++ b/test/run-fail/mem/free-stack-alloca.c
@@ -1,0 +1,11 @@
+
+#include "caffeine.h"
+#include <stdlib.h>
+
+void test(unsigned x) {
+  unsigned mem[32];
+  mem[5] = x;
+  free(mem);
+
+  caffeine_assert(mem[5] == x);
+}
--- a/test/run-fail/mem/free-stack-alloca.c
+++ b/test/run-fail/mem/free-stack-alloca.c
@@ -7,5 +7,6 @@ void test(unsigned x) {
   mem[5] = x;
   free(mem);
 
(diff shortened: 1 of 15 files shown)


test/run-fail/mem/free-stack-alloca.c
```diff
@@ -0,0 +1,11 @@
+
+#include "caffeine.h"
+#include <stdlib.h>
+
+void test(unsigned x) {
+  unsigned mem[32];
+  mem[5] = x;
+  free(mem);
+
+  caffeine_assert(mem[5] == x);
```
mishazharov, line 10: nit: This assertion shouldn't be needed since it should fail on the free method

mishazharov (approved):
Nice, LGTM!

test/run-fail/mem/free-stack-alloca.c
Phantomical: This is needed to prevent clang from optimizing the call to free away. I've added a comment instead.

Issue #96, opened 2021-01-10
Implement constant-folding for floating-point operations


Phantomical:
Note that many of the identities for integers don't necessarily apply to floats. Examples:
- `x * 0` is NaN if x is NaN or infinity
- anything involving NaN really
- there are specific rules for combining the mantissa bits when multiple NaNs are involved

So for this issue constant folding should only take place if all operands are constants.

PR #132, opened 2021-01-20
Add const folding for float add and sub

mishazharov:
This PR is 1/2 for implementing constant folding for floating point expressions.

Part of #96

### DIFF
--- a/src/IR/Operation.cpp
+++ b/src/IR/Operation.cpp
@@ -695,11 +695,69 @@ ref<Operation> BinaryOp::CreateAShr(const ref<Operation>& lhs,
   return Create(Opcode::AShr, lhs, rhs);
 }
 
-DECL_BINOP_CREATE(FAdd, ASSERT_FP);
-DECL_BINOP_CREATE(FSub, ASSERT_FP);
-DECL_BINOP_CREATE(FMul, ASSERT_FP);
-DECL_BINOP_CREATE(FDiv, ASSERT_FP);
-DECL_BINOP_CREATE(FRem, ASSERT_FP);
+#define PROPAGATE_NANS(lhs, rhs)                                               \
+  if (EXTRACT_FLOAT_TO(*lhs, constant)) {                                      \
+    if (constant->isNaN()) {                                                   \
+      return lhs;                                                              \
+    }                                                                          \
+  }
+
+ref<Operation> BinaryOp::CreateFAdd(const ref<Operation>& lhs,
+                                    const ref<Operation>& rhs) {
+  CAFFEINE_ASSERT(lhs, "lhs was null");
+  CAFFEINE_ASSERT(rhs, "rhs was null");
+  ASSERT_FP(lhs);
+  ASSERT_FP(rhs);
+
+  if (const auto* var_name =
+          llvm::dyn_cast<caffeine::ConstantFloat>(lhs.get())) {}
+
+  // PROPAGATE_NANS(lhs, rhs);
+
+#ifdef CAFFEINE_IMPLICIT_CONSTANT_FOLDING
+  // if () {
+
+  // }
+#endif
+
+  return Create(Opcode::FAdd, lhs, rhs);
+}
+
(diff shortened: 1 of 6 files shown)


src/Interpreter/CtxConstEval.cpp
```diff
@@ -199,6 +270,12 @@ static ContextValue evaluate_expr(Context* ctx, llvm::ConstantExpr* expr) {
   os.flush();
 
   CAFFEINE_UNIMPLEMENTED(s);
+
+#undef UNARY_OP
+#undef BINARY_OP
+#undef CAST_OP
+#undef UNARY_NAN_PROPAGATE
+#undef BINARY_NAN_PROPAGATE
```
mishazharov, line 278: This is optional since the macros are in a source file but I thought it was nice anyway

Phantomical:
Shouldn't this be changing the `CreateXXX` functions in Operation.cpp? Constant folding is also needed outside of evaluating globals

mishazharov:
> Shouldn't this be changing the `CreateXXX` functions in Operation.cpp? Constant folding is also needed outside of evaluating globals.

Yup I flubbed it. Will fix

src/IR/Operation.cpp
```diff
@@ -695,11 +698,106 @@ ref<Operation> BinaryOp::CreateAShr(const ref<Operation>& lhs,
...
+  ASSERT_FP(lhs);
+  ASSERT_FP(rhs);
+
+#ifdef CAFFEINE_IMPLICIT_CONSTANT_FOLDING
+  const auto* lhs_float = llvm::dyn_cast<caffeine::ConstantFloat>(lhs.get());
+  const auto* rhs_float = llvm::dyn_cast<caffeine::ConstantFloat>(rhs.get());
+
+  if (lhs_float && rhs_float) {
+    return caffeine::ConstantFloat::Create(
+        fmod(lhs_float->value().convertToFloat(),
```
Phantomical, line 794: This doesn't work if the value stored in the APFloat isn't a float. Isn't there a method on APFloat to do this properly?

Phantomical (changes requested):
One last thing: it's basically never correct to convert an APFloat back to a float or double for calculation purposes since that'll give wrong values for other float sizes (what about fp80 or fp128?)

src/IR/Operation.h
```diff
@@ -51,22 +51,25 @@ inline bool constant_int_compare(ICmpOpcode cmp, const llvm::APInt& lhs,
...
-    return lhs_float != rhs_float;
+    return lhs.compare(rhs) != llvm::APFloat::cmpEqual;
   case FCmpOpcode::GE:
-    return lhs_float >= rhs_float;
+    res = lhs.compare(rhs);
+    return res == llvm::APFloat::cmpGreaterThan ||
+           res == llvm::APFloat::cmpEqual;
   case FCmpOpcode::GT:
-    return lhs_float > rhs_float;
+    return lhs.compare(rhs) != llvm::APFloat::cmpGreaterThan;
```
Phantomical, line 67: Isn't this backwards? Shouldn't it be `lhs.compare(rhs) == llvm::APFloat::cmpGreaterThan`?

src/IR/Operation.h
```diff
@@ -51,22 +51,25 @@ inline bool constant_int_compare(ICmpOpcode cmp, const llvm::APInt& lhs,
...
   case FCmpOpcode::LE:
-    return lhs_float <= rhs_float;
+    res = lhs.compare(rhs);
+    return res == llvm::APFloat::cmpLessThan || res == llvm::APFloat::cmpEqual;
   case FCmpOpcode::LT:
-    return lhs_float < rhs_float;
+    return lhs.compare(rhs) != llvm::APFloat::cmpLessThan;
```
Phantomical, line 72: Same thing here.

Phantomical (approved):
Approved provided that you fix the last two comments :)

src/IR/Operation.h
mishazharov: Nice catch, copy pasta strikes again

PR #145, opened 2021-01-22
Miscellaneous refactoring

Phantomical:
It takes a lot of typing to create a binary operation with a constant integer operand. This PR adds methods that remove some of the useless boilerplate.

Specifically, for each `CreateXXX` method in `BinaryOp` there are now two more overloads that allow passing an integer in on each side. I've also gone through the codebase and simplified all the places where the new methods could apply.

### DIFF
--- a/src/Interpreter/Interpreter.cpp
+++ b/src/Interpreter/Interpreter.cpp
@@ -28,8 +28,7 @@ auto zip(R1& range1, R2& range2) {
          });
 }
 
-#define WRAP_FUNC(func) \
-  [](const auto&... args) { return (func)(args...); }
+#define WRAP_FUNC(func) [](const auto&... args) { return (func)(args...); }
 
 Interpreter::Interpreter(Executor* queue, Context* ctx, FailureLogger* logger,
                          const InterpreterOptions& options)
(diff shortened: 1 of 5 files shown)


Skrinikov (approved):
🚢 🇮🇹

src/Interpreter/Interpreter.cpp
```diff
@@ -28,6 +28,8 @@ auto zip(R1& range1, R2& range2) {
          });
 }
 
+#define WRAP_FUNC(func) [](const auto&... args) { return (func)(args...); }
+
```
mishazharov, line 32: Is `WRAP_FUNC` necessary? It's starting to feel pretty verbose

mishazharov (approved):

src/Interpreter/Interpreter.cpp
Phantomical: Unfortunately, yes. The problem is that now that there's multiple overloads of, say, `BinaryOp::CreateAdd` it can't figure out which one to pass as the function pointer.

We could work around this by doing something like `#define transform(func, ...) transform(WRAP_FUNC(func), __VA_ARGS__)` but that felt pretty magical so I'd rather now.

PR #146, opened 2021-01-22
Check read width when validating pointers

### LINKED ISSUE #103: Change `MemHeap::check_valid` to take read width into account.
Right now it just checks whether the pointer is within any allocation but for each allocation it should really also be checking:
- Whether the range `[ptr, ptr + size)` is within the allocation
- That `ptr + size` doesn't wrap around the address space.

Phantomical:
Previously a read that was only partially within an allocation wouldn't trigger an error but would instead just silently delete that branch. That was incorrect and it now will properly pass a test failure.

To make it easier to keep track of regression tests like this I've added a `test/regressions` folder with tests that are meant to be named in a defined format that makes it easier to keep look at the file name and find the relevant issue.

Closes #103

### DIFF
--- a/include/caffeine/Memory/MemHeap.h
+++ b/include/caffeine/Memory/MemHeap.h
@@ -208,7 +208,7 @@ public:
    * the absolute value with those of all existant allocations and returns an
    * assertion that the pointer points within one of them.
    */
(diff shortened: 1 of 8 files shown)


Skrinikov (approved):
looks good

PR #147, opened 2021-01-23
[1/2] Optimize internal representation of known-size symbolic arrays

Phantomical:
Our current representation of such arrays is really slow. If we know the length of the array then we can use a more efficient representation that avoids making array value lookups go through z3 a lot of the time.

This PR adds a new data structure, called a `PersistentArray`. It's an array that shares the underlying data between multiple instances but also allows those instances to accept writes without the writes being shared.

The majority of the complicated implementation details happen within the `PersistentArray::Node` class, as well as the `get` and `set` methods. A lot of the rest is just boilerplate.

### DIFF
--- a/include/caffeine/ADT/PersistentArray.h
+++ b/include/caffeine/ADT/PersistentArray.h
@@ -55,6 +55,8 @@ namespace detail {
  * This data structure is an implementation of the persistant array data
  * structure as described in "A Persistent Union-Find Data Structure" by
  * Conchon, S. and Filliatre, J.
+ * 
+ * The paper itself can be found at https://www.lri.fr/~filliatr/puf/
  *
  * Efficiency Notes
  * ================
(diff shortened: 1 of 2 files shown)


include/caffeine/ADT/PersistentArray.h
```diff
@@ -0,0 +1,442 @@
...
+ * This data structure is an implementation of the persistant array data
+ * structure as described in "A Persistent Union-Find Data Structure" by
+ * Conchon, S. and Filliatre, J.
+ *
+ * The paper itself can be found at https://www.lri.fr/~filliatr/puf/
+ *
+ * Efficiency Notes
+ * ================
+ * Internally, a persistent array is represented as a base array + a tree of
+ * changes to the base array. When a store is doing the tree is rerooted to
```
mishazharov, line 64: > When a store is doing the tree

Sounds clunky

include/caffeine/ADT/PersistentArray.h
```diff
@@ -0,0 +1,442 @@
...
+  };
+
+  template <typename... Ts>
+  using first_of_t = typename first_of<Ts...>::type;
+
+  /**
+   * Template to check whether the type T has an operator== overload for itself.
+   */
+  template <typename T>
+  class is_comparable {
```
mishazharov, line 30: Is this used anywhere?

mishazharov (approved):
LGTM!

PR #148, opened 2021-01-23
Add script to generate random c++ tests using yarpgen

### LINKED ISSUE #114: Investigate using C++ randomized testers
One example that I spotted today was https://github.com/intel/yarpgen. This would be a good starting place for more complete test coverage.

KLEE did something similar to this so it might be worth looking at what they did.

Phantomical:
None of the generated tests currently use anything symbolic but they're useful for finding unsupported features. I haven't included any of the tests here since I'm not really sure if we want to add a bunch of autogenerated tests to the commit history.

I've also included the fixes necessary to get C++ code to compile when passed in to `add_llvm_ir_library`.

Closes #114


cmake/LLVMIRUtils.cmake
```diff
@@ -47,15 +47,22 @@ set_if_unset(LLVM_LINK "${LLVM_TOOLS_BINARY_DIR}/llvm-link")
...
+set(CMAKE_LLVM_CXX_IMPLICIT_LINK_LIBRARIES "")
+set(CMAKE_LLVM_CXX_IMPLICIT_LINK_DIRECTORIES "")
 
 set(
   CMAKE_BITCODE_CREATE_SHARED_LIBRARY
   "${CMAKE_COMMAND} \"-DLINKER=${LLVM_LINK}\" \"-DFLAGS=<LINK_FLAGS> -S\" \"-DTARGET=<TARGET>\" \"-DLINK_LIBRARIES=<LINK_LIBRARIES>\" \"-DOBJECTS=<OBJECTS>\" -P \"${CLANG_LINK_SCRIPT}\""
 )
+# set(CMAKE_BITCODE_LINK_LIBRARY_FILE_FLAG "")
+# set(CMAKE_BITCODE_LINK_LIBRARY_FLAG "")
+set(CMAKE_BITCODE_IMPLICIT_LINK_LIBRARIRES "")
```
mishazharov, line 65: It might be a good idea to explain why it's commented or remove the comments

mishazharov (approved):

cmake/LLVMIRUtils.cmake
Phantomical: Whoops. Yeah that's left over from some experimentation I was doing. Will remove.

PR #149, opened 2021-01-23
Implement non-ConstantData globals

Phantomical:
This fixes one of the issues discovered while running the test cases generated by #148. Specifically we don't support globals that don't derive from `ConstantData`.

Here's the summary of the changes that were added to make this work:
- Added a pure virtual `ArrayBase` operation so that it's possible to get the size of an array generically. 
- Implemented global evaluation by evaluating the constant itself and then storing writing it to a data array.

I haven't implemented support for vector stores yet since what that will actually entail hasn't been worked out yet. I've added a simple test case but most of the behaviour here makes use of other, already tested, work.

### DIFF
--- a/include/caffeine/IR/Operation.h
+++ b/include/caffeine/IR/Operation.h
@@ -316,7 +316,8 @@ public:
   Operation& operator=(const Operation& op);
   Operation& operator=(Operation&& op) noexcept;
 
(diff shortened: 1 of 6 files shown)


mishazharov (approved):
LGTM

Issue #92, opened 2021-01-09
Verify run-test failures actually fail with returned test cases


Phantomical:
Currently run-fail tests just check to see whether caffeine returns a failure. However, we should also be checking that the test case returned by caffeine actually causes a failure.

There's a few levels of this:
1. Check that the returned output matches a specified one. This only works for tests where there's only 1 possible set of inputs that cause a failure.
2. Interpret/execute the code to find the failure. This will need an interpreter that checks all the error conditions so it is likely to be much more work.

PR #152, opened 2021-01-25
Add const lookup to Context

mishazharov:
This PR adds a `const` `lookup` method to `Context` which is necessary for #92 in order to check values in the `FailureHandler`


include/caffeine/Interpreter/CtxConstEval.h
```diff
@@ -0,0 +1,257 @@
...
+#include <llvm/IR/Instruction.h>
+#include <llvm/IR/Module.h>
+#include <llvm/Support/raw_ostream.h>
+
+using llvm::Instruction;
+
+namespace caffeine {
+
+template <typename ContextParamTy>
+static std::optional<ContextValue> evaluate_expr(ContextParamTy ctx,
```
mishazharov, line 24: Mostly copy pasted from `CtxConstEval.cpp` with macro changes

cmake/testutils.cmake
```diff
@@ -31,9 +31,9 @@ function(declare_test TEST_NAME_OUT test)
   should_skip_test(should_skip "${test}")
 
   string(REGEX REPLACE "\\\\|/" "_" test_target "${test_name}")
-  
+
```
mishazharov, line 34: vscode got rid of some whitespace so ¯\\_(ツ)_/¯

src/bin/caffeine.cpp
```diff
@@ -25,12 +25,21 @@ using caffeine::Interpreter;
...
+  CountingFailureLogger(std::ostream& os, llvm::Function* func)
+      : PrintingFailureLogger(os), func{func} {}
+
   void log_failure(const caffeine::Model& model, const caffeine::Context& ctx,
                    const caffeine::Failure& failure) override {
     num_failures += 1;
+
+    for (auto& arg : func->args()) {
+      ctx.lookup(&arg);
+    }
```
mishazharov, line 41: Empty lookups which do nothing right now

include/caffeine/Interpreter/Context.h
```diff
@@ -133,8 +136,11 @@ class Context {
    * Not a constant method since it may have to create allocations and
    * modify the registered globals table.
```
Phantomical, line 137: I think we need a second set of docs for the `const` version of `evaluate_constant` to communicate just how it differs.

Since it actually has different behaviour I think it's worth putting it under a different name? It's going to be really confusing to use this API since the const and non-const perform similar, but not quite the same, operations.

include/caffeine/Interpreter/Context.h
```diff
@@ -107,6 +109,7 @@ class Context {
    * This method should be preferred over directly looking up variables in the
    * stack frame as it properly handles global constants.
    */
+  std::optional<ContextValue> lookup(llvm::Value* value) const;
```
Phantomical, line 112: Same comment as below (on both counts). Maybe `lookup_const` would work as a name?

Phantomical (changes requested):
Lots of comments here, sorry. I think the general approach is right but there's some issues:
- Can you move `evaluate_expr` back into `CtxConstEval.cpp`. It's an internal helper function that shouldn't be accessible to external users and moving it makes the diff much harder to review (IMO) unnecessarily.
- Some of the other helpers also need to be turned into templates, I've marked which ones I noted would need this.
- Having const and non-const overloads of a function that do slightly different things (beyond just const/non-const) is quite easy to screw up for users of the API. Maybe rename the constant overloads to distinguish them? The const overloads also need docs to explain what they're doing and how they're different from the non-const ones.

In short - the functionality looks correct (+/- some template overload issues). Most of my issues here are around code layout and naming.

src/Interpreter/CtxConstEval.cpp
```diff
@@ -54,15 +67,27 @@ static ContextValue evaluate_const_vector(Context* ctx,
...
+static std::optional<ContextValue> evaluate_expr(ContextType ctx,
+                                                 llvm::ConstantExpr* expr) {
+  static_assert(
+      std::is_same_v<std::remove_const_t<std::remove_pointer_t<ContextType>>,
+                     Context>,
+      "ContextType must be a Context pointer");
+
 #define OPERAND(expr, num)                                                     \
-  evaluate(ctx, llvm::cast<llvm::Constant>((expr)->getOperand(num)))
+  *evaluate(ctx, llvm::cast<llvm::Constant>((expr)->getOperand(num)))
```
mishazharov, line 90: This is a bug since the optional result isn't checked and is just dereferenced right away

src/bin/caffeine.cpp
```diff
@@ -25,12 +25,21 @@ using caffeine::Interpreter;
 class CountingFailureLogger : public caffeine::PrintingFailureLogger {
 public:
   uint64_t num_failures = 0;
+  llvm::Function* func;
 
   using caffeine::PrintingFailureLogger::PrintingFailureLogger;
```
Phantomical, line 30: You'll want to remove this. If CountingFailureLogger is constructed with the inherited constructor then `func` won't get initialized.

Phantomical (approved):
One thing left. LGTM otherwise!

PR #157, opened 2021-01-28
Implement expression-tree matching utilities

Phantomical:
Pretty soon we're going to want to perform various query optimizations. As part of this we'll need an easy way to check if an expression is an expected expression tree. This PR is the first part of what's needed for that. It adds a (somewhat) complete set of patterns along with methods to use them.

To see how this code is supposed to be used check out the docs on `matches` and `matches_anywhere`.

### DIFF
--- a/include/caffeine/IR/Matching.h
+++ b/include/caffeine/IR/Matching.h
@@ -38,8 +38,11 @@ namespace matching {
    * and have first and second be set to the expression nodes that were matched
    * against. However, ref<Operation> doesn't implement the protocol (and you'd
    * need a reference anyway). To work around this matchers use the MatcherImpl
-   * type internally. This is a template which resolves to T unless T is a
-   * ref<Operation> in which case it resolves to RefOperationMatcher.
+   * type internally. This is a template which resolves to T unless T is one of
+   * a few custom types in which case it resolves to custom matcher.
+   *
+   * Currently the only special case is RefOperationMatcher if T is
+   * ref<Operation> or ref<Operation>*.
    *
    * For an example of how this is used check out UnaryOpMatcher. Custom
    * matchers should be implemented along the same lines.
--- a/include/caffeine/IR/Matching.h
+++ b/include/caffeine/IR/Matching.h
@@ -7,19 +7,47 @@ namespace caffeine {
 namespace matching {
 
   /**
-   * Matching protocol:
-   *
+   * Writing a Matcher
+   * =================
    * Every matching object exposes two methods:
-   *  - bool matches(<ast node>) const
-   *  - void on_match(<ast node>) const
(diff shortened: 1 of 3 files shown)


include/caffeine/IR/Matching.h
```diff
@@ -0,0 +1,424 @@
...
+
+  public:
+    RefOperationMatcher(ref<Operation>& output) : output(&output) {}
+    RefOperationMatcher(ref<Operation>* output) : output(output) {}
+
+    bool matches(const ref<Operation>&) const {
+      return true;
+    }
+    void on_match(const ref<Operation>& op) const {
+      *output = op;
```
mishazharov, line 67: Any concerns about `nullptr`?

mishazharov (approved):
some nits but otherwise lgtm!

include/caffeine/IR/Matching.h
```diff
@@ -0,0 +1,424 @@
...
+   * Currently the only special case is RefOperationMatcher if T is
+   * ref<Operation> or ref<Operation>*.
+   *
+   * For an example of how this is used check out UnaryOpMatcher. Custom
+   * matchers should be implemented along the same lines.
+   */
+
+  /**
+   * Matcher to capture an expression tree node.
+   */
```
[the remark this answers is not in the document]
Phantomical: I initially tried that but found it too verbose. Instead I just made it so that you can pass a reference to a `ref<Operation>` in and it'll implicitly match anything. This class is an implementation detail for that.

`Any` is still around because sometimes you don't want to capture anything.

include/caffeine/IR/Matching.h
Phantomical: This struct is really an implementation detail and should basically never be used explicitly. If someone passes in `nullptr` then we'll just crash. I think that's more or less ok.

include/caffeine/IR/Matching.h
```diff
@@ -0,0 +1,424 @@
...
+    matcher.on_match(op);
+  return is_match;
+}
+template <typename Matcher>
+bool matches(const Assertion& assertion, const Matcher& matcher) {
+  return matches(assertion.value(), matcher);
+}
+
+/**
+ * Match anywhere in an expression.
```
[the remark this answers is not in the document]
Phantomical: I don't really find either of those clearer. I don't think we'll ever have a `matches_everywhere` method (it would be useless semantically) so I'd rather just stick with anywhere since it's the clearest name IMO.

include/caffeine/IR/Matching.h
mishazharov: In that case maybe it's worth it to move it into the `detail` namespace? Honestly not too worried about it feel free to merge

PR #177, opened 2021-01-29
Implement canonicalizing solver

Phantomical:
This is a solver that attempts to convert expressions to a canonical form. We haven't really defined what we want out of a canonical form yet so all the solver does right now is eliminate duplicate assertions and remove unevaluated constant subexpressions.

### DIFF
--- a/src/IR/Transforms/canonicalize.cpp
+++ b/src/IR/Transforms/canonicalize.cpp
@@ -81,7 +81,6 @@ void canonicalize(std::vector<Assertion>& assertions) {
           [](const Assertion& a) { return a.is_constant_value(false); })) {
     assertions.clear();
     assertions.push_back(Assertion(ConstantInt::Create(false)));
-    return;
   }
 }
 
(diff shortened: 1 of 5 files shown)


src/IR/Transforms/canonicalize.cpp
```diff
@@ -0,0 +1,88 @@
...
+  assertions.erase(it, assertions.end());
+
+  // The canonical form for an assertion that has a false literal is just an
+  // array with a single constant-false assertion.
+  if (std::any_of(
+          std::begin(assertions), std::end(assertions),
+          [](const Assertion& a) { return a.is_constant_value(false); })) {
+    assertions.clear();
+    assertions.push_back(Assertion(ConstantInt::Create(false)));
+    return;
```
Skrinikov, line 84: Is this return necessary?

Phantomical: Fixed.

mishazharov (approved):

Issue #173, opened 2021-01-29
Modify Model class


mishazharov:
The Model superclass evaluate expressions, and its subclasses just take care of lookups of constants in various solvers. The implications of this is that `evaluate` doesn't need to be `virtual` and `lookup` should be protected.

Also add some documentation for class to explain that.

Finally we want an overload for `evaluate(ContextValue)` for easier access.

PR #178, opened 2021-02-01
Refactor the model class

mishazharov:
Copy pasted from #173:

The Model superclass evaluate expressions, and its subclasses just take care of lookups of constants in various solvers. The implications of this is that `evaluate` doesn't need to be `virtual` and `lookup` should be protected.

Also add some documentation for class to explain that.

Finally we want an overload for `evaluate(ContextValue)` for easier access.

### DIFF
--- a/include/caffeine/IR/Value.h
+++ b/include/caffeine/IR/Value.h
@@ -46,11 +46,9 @@ private:
     T data;
     uint32_t index_bitwidth;
 
-    ArrayData(const T& data) : data(data), index_bitwidth(0){};
-    ArrayData(T&& data) : data(data), index_bitwidth(0){};
-    ArrayData(const T& data, uint32_t idx_width)
+    ArrayData(const T& data, uint32_t idx_width = 0)
         : data(data), index_bitwidth(idx_width){};
-    ArrayData(T&& data, uint32_t idx_width)
+    ArrayData(T&& data, uint32_t idx_width = 0)
         : data(data), index_bitwidth(idx_width){};
   };
 
--- a/include/caffeine/IR/Value.h
+++ b/include/caffeine/IR/Value.h
@@ -37,19 +37,23 @@ class ConstantFloat;
 class Value {
 public:
   // Note: These correspond to the variant indices.
-  enum Kind { Empty, Int, Float, Array };
+  enum Kind { Empty, Int, Float, Array, NestedArray };
 
 private:
+  template <typename T>
   struct ArrayData {
-    SharedArray data;
+    T data;
     uint32_t index_bitwidth;
 
-    ArrayData(const SharedArray& data, uint32_t idx_width);
-    ArrayData(SharedArray&& data, uint32_t idx_width);
+    ArrayData(const T& data, uint32_t idx_width)
+        : data(data), index_bitwidth(idx_width){};
+    ArrayData(T&& data, uint32_t idx_width)
+        : data(data), index_bitwidth(idx_width){};
   };
 
   using Inner =
-      std::variant<std::monostate, llvm::APInt, llvm::APFloat, ArrayData>;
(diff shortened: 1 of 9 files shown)


src/Solver/Solver.cpp
```diff
@@ -121,6 +122,31 @@ Value Model::evaluate(const Operation& expr) const {
...
+
+    std::transform(vec.begin(), vec.end(), nested_arr.begin(),
+                   [&](const auto& i) -> Value { return evaluate(i, ctx); });
+
+    // If it's an empty nested array, then the bitwidth is 0 since we can't
+    // really determine it from anything else
+    return Value(std::move(nested_arr),
+                 vec.size() == 0
+                     ? Type::int_ty(0)
+                     : Type::int_ty(nested_arr[0].type().bitwidth()));
```
Phantomical, line 144: This looks weird to me and suggests there's something wrong with the data model here.

include/caffeine/IR/Value.h
```diff
@@ -37,19 +37,23 @@ class ConstantFloat;
 class Value {
 public:
   // Note: These correspond to the variant indices.
-  enum Kind { Empty, Int, Float, Array };
+  enum Kind { Empty, Int, Float, Array, NestedArray };
```
Phantomical, line 40: Would a better name for this be `Vector`?

This is basically a nit though, feel free to ignore.

Phantomical (commented):
Thinking about it a bit more, I don't think LLVM vectors really have an index bitwidth associated with them? They're always accessed using (effectively) untyped integers so I don't think it makes sense.

Here's my alternate suggestion:
1. Add a new vector type to `Type`.
2. Change the vector `Value` constructor to only take `std::vector`. (The internal implementation can do whatever).

Everything else LGTM. If you feel the current approach is ok then I'll approve. I just think that if the design around this leads to awkward things like whats going on in the comment below then there's something wrong.

include/caffeine/IR/Value.h
mishazharov: In llvm vectors are basically list that contain only primitives, so calling this a vector would be a bit confusing.

include/caffeine/IR/Value.h
```diff
@@ -37,19 +38,25 @@ class ConstantFloat;
...
+  template <typename T>
   struct ArrayData {
-    SharedArray data;
+    T data;
     uint32_t index_bitwidth;
 
-    ArrayData(const SharedArray& data, uint32_t idx_width);
-    ArrayData(SharedArray&& data, uint32_t idx_width);
+    ArrayData(const T& data) : data(data), index_bitwidth(0){};
+    ArrayData(T&& data) : data(data), index_bitwidth(0){};
```
Phantomical, line 50: Nit: You can avoid having to write these by having `idx_width = 0` as a default parameter in the ctors below.

src/Solver/Z3Solver.cpp
```diff
@@ -142,6 +142,8 @@ static z3::sort type_to_sort(z3::context& ctx, const Type& type) {
     CAFFEINE_ABORT("Cannot make symbolic pointer constants");
   case Type::FunctionPointer:
     CAFFEINE_ABORT("Cannot make symbolic function constants");
+  case Type::Vector:
+    CAFFEINE_ABORT("Symbolic vectors are unimplemented");
```
Phantomical, line 146: Nit: We should never have symbolic vectors (this is one of the jobs of the interpreter) so maybe this should say something like `"cannot make symbolic vector constants"`?

Phantomical (approved):
LGTM! Some minor nits but they're really not that important, feel free to ignore them.

include/caffeine/IR/Value.h
mishazharov: derp brain fart, nice catch

PR #199, opened 2021-02-15
Implement Symbol type for constant names

Phantomical:
This is meant to make it easier to write code that just wants to deal with identifier of a constant without caring if it's a string or a number. I haven't changed any of the downstream callsites yet but new code should use `Symbol` if it doesn't need to care about the string vs number divide.


mishazharov (approved):
ah we're slowly converging on duck typing ;)

PR #201, opened 2021-02-15
Remove special-casing for FixedArray

Phantomical:
In current master FixedArray doesn't behave like any of the other operation classes. This PR changes it to align with the way other classes are used and removes special casing elsewhere within the code.

Detailed list of changes:
- Add code to copy the class vtable pointer when an `Operation` instance is copied
- Override a number of member functions within `Operation` so that they work when the derived type is a `FixedArray` instance.
- Ensure that the vtable is copied when calling `with_new_operands`
- Expose some new methods on `PersistentArray` to support the operations that `FixedArray` needs

See the docs on `CopyVTable` for some of the restrictions on using it properly.

### DIFF
--- a/include/caffeine/Support/CopyVTable.h
+++ b/include/caffeine/Support/CopyVTable.h
@@ -1,0 +1,54 @@
+#pragma once
+
+namespace caffeine {
+
+/**
+ * Copy the vtable of all derived classes.
+ *
+ * Usage
+ * =====
+ * Have whatever class is at the base of your object hierarchy derive
+ * CopyVTable.
+ *
+ * Explanation
+ * ===========
+ * By default, when you copy a potentially-derived instance of a base class in
+ * C++ it doesn't copy the vtable of the object along with the rest of it (as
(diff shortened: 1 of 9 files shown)


include/caffeine/Support/CopyVTable.h
```diff
@@ -0,0 +1,54 @@
...
+  CopyVTable(const CopyVTable&) noexcept;
+  CopyVTable(CopyVTable&&) noexcept;
+
+  CopyVTable& operator=(const CopyVTable&) noexcept;
+  CopyVTable& operator=(CopyVTable&&) noexcept;
+
+  void copy_vtable(const CopyVTable& vtable) noexcept;
+};
+
+} // namespace caffeine
```
Skrinikov, line 54: fmt

Skrinikov (approved):

include/caffeine/Support/CopyVTable.h
Phantomical: This is formatted? (Otherwise it wouldn't have passed ci)

PR #204, opened 2021-02-20
Update solver API to use the Symbol type

Phantomical:
Now that we have a common type for constant identifiers we should be using it anywhere we would previously take a symbol name or number. This updates `Model::lookup` to use `Symbol` (and also adds a size parameter to support the upcoming ConstantArray operation).


mishazharov (approved):
LGTM but the build needs to be fixed

PR #206, opened 2021-02-20
Implement caffeine_make_symbolic

### LINKED ISSUE #134: Implement caffeine_make_symbolic
API documentation:
```
/**
 * Makes the memory range pointed to by mem symbolic for at least size bytes with name name.
 *
 * name must not be symbolic.
 */
caffeine_make_symbolic(void* mem, size_t size, const char* name);
```

Phantomical:
This PR implements the `caffeine_make_symbolic` builtin. See the docs on the builtin for what it actually does.

I have also added a single test case to help verify printing and whether it actually works.

Closes #134

### DIFF
--- a/libraries/builtins/make_symbolic.c
+++ b/libraries/builtins/make_symbolic.c
@@ -1,0 +1,10 @@
+
+#include "caffeine-builtins.h"
+#include <string.h>
+
+void caffeine_make_symbolic(void* mem, size_t size, const char* name) {
+  caffeine_assert(!!name);
+
+  void* symbolic = caffeine_builtin_symbolic_alloca(size, name);
+  memcpy(mem, symbolic, size);
+}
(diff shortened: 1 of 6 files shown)


libraries/builtins/make_symbolic.c
```diff
@@ -0,0 +1,10 @@
+
+#include "caffeine-builtins.h"
+#include <string.h>
+
+void caffeine_make_symbolic(void* mem, size_t size, const char* name) {
+  caffeine_assert(!!name);
```
mishazharov, line 6: What does this do?

mishazharov (approved):

libraries/builtins/make_symbolic.c
Phantomical: It asserts that `name` isn't null.

PR #210, opened 2021-02-24
Add private utility method to print an expression while in a debugger

Phantomical:
This is a private method which prints a representation of the expression to stdout. It's meant to be used as an aid while debugging so I've left it as a private method.

Note that this only really works when caffeine is built as a shared library as enabled by #209.


mishazharov (approved):
What stops it from working when it's a static lib?

Phantomical:
> 
> 
> What stops it from working when it's a static lib?

The linker will remove the method since it (rightly) sees that it's not callable anywhere.

PR #212, opened 2021-02-24
Fix use of CopyVTable within Operation

Phantomical:
As it turns out, the copy constructor for CopyVTable runs before the derived class actually initializes the derived vtable. This means that as soon as we actually start copying Operations around the vtable was not being copied around with them. This showed up once the code actually started copying operations around and not just moving refcounted pointers around.

The fix for this is to have Operation call copy_vtable within its ctors. I've prevented the same issue from (hopefully) occurring again by deleting the copy/move ctors of CopyVTable and documenting this footgun in the class documentation.

### DIFF
--- a/include/caffeine/Support/CopyVTable.h
+++ b/include/caffeine/Support/CopyVTable.h
@@ -20,7 +20,11 @@ namespace caffeine {
  *
  * However, if you know that all dervied classes will have exactly the same size
  * as the base class and you want to have the classes vtable copied along with
- * them then you should take use this class.
+ * them then you should use this class. To actually make use of this class you
+ * will have to call copy_vtable within the copy/move ctors/assigment operators.
+ * This is because the compiler overwrites the vtable during construction of the
+ * derived operation so copy_vtable needs to be called after the object is
+ * initialized (i.e. within the body of the constructor).
  *
  * Limitations
  * ===========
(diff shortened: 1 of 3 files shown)


include/caffeine/Support/CopyVTable.h
```diff
@@ -20,7 +20,11 @@ namespace caffeine {
...
  * as the base class and you want to have the classes vtable copied along with
- * them then you should take use this class.
+ * them then you should use this class. To actually make use of this class you
+ * will have to call copy_vtable within the copy/move ctors/assigment operators.
+ * This is because the compiler overwrites the vtable during construction of the
+ * derived operation so copy_vtable needs to be called after the object is
+ * initialized (i.e. within the body of the constructor).
  *
  * Limitations
  * ===========
```
mishazharov, line 30: Unrelated to this PR but is there anything under the limitations header?

mishazharov (approved):

include/caffeine/Support/CopyVTable.h
Phantomical: Yup. It looks like it's just not shown by default in the github UI. There's a thing at the side (just under this comment for me) that you can click to expand and show the in-between text.

mishazharov: I'm blind

PR #215, opened 2021-02-24
Implement expression simplification solver

### LINKED ISSUE #161: Create instruction simplification solver pass

Phantomical:
This PR implements the start of a generic expression simplification solver. Currently all it is able to do is run the decomposition transform and replace with constants when it finds an `x == <const>` constraint within the assertions. In the future this should be expanded as needed.

Closes #161.

### DIFF
--- a/src/IR/Transforms/simplify.cpp
+++ b/src/IR/Transforms/simplify.cpp
@@ -7,8 +7,8 @@ namespace m = caffeine::matching;
 
 namespace {
 
-  bool is_equality_expr(const Assertion& assertion, ref<Operation>& constant,
-                        ref<Operation>& value) {
+  bool is_equality_expr(const Assertion& assertion, OpRef& constant,
+                        OpRef& value) {
     using namespace caffeine::matching;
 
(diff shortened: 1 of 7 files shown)


src/IR/Transforms/simplify.cpp
```diff
@@ -0,0 +1,60 @@
...
+    if (!is_equality_expr(assertions[i], constant_, value))
+      continue;
+
+    const auto* constant = llvm::cast<Constant>(constant_.get());
+
+    for (size_t j = 0; j < assertions.size(); ++j) {
+      if (j == i)
+        continue;
+
+      auto changed = rebuild(assertions[i].value(), [&](const OpRef& op) {
```
mishazharov, line 44: Should `assertions[i]` be `assertions[j]` on this line? Otherwise I'm not too sure what's going on here (the assertions are being rebuilt `j` times?

Phantomical: Whoops - good catch. This is probably telling me I should have test cases but I'm going to procrastinate on that until later.

mishazharov (approved):

PR #218, opened 2021-02-28
Start building out serialization infrastructure

mishazharov:
This PR adds:

1) A class for mapping `llvm::Value`'s to IDs
2) CMake modifications to use capnp


src/Serialization/ValueIdMap.cpp
```diff
@@ -0,0 +1,33 @@
...
+    curr++;
+  }
+};
+
+std::optional<ValueId> ValueIdMap::getValueId(llvm::Value* val) {
+  if (mapping_.find(val) == mapping_.end()) {
+    return std::nullopt;
+  }
+
+  return mapping_.at(val);
```
Phantomical, line 30: This code will perform the lookup twice. Maybe do this instead?
```suggestion
  auto it = mapping_.find(val);
  if (it == mapping.end()) {
    return std::nullopt;
  }
  
  return it->second;
```

src/Protos/operation.capnp
```diff
@@ -0,0 +1,9 @@
+@0xf54292a58fc34e9a;
+
+
+struct Symbol {
```
Phantomical, line 4: Can you format this with 2 spaces like the rest of our code? It's best to be consistent.

include/caffeine/Serialization/ValueIdMap.h
```diff
@@ -0,0 +1,34 @@
+#ifndef CAFFEINE_DIST_SERIALIZATION_H
+#define CAFFEINE_DIST_SERIALIZATION_H
```
Phantomical, line 2: Small thing: we can probably just use #pragma once here. All the compilers that can compile our code support it.

Phantomical (approved):
LGMT! Just some small nits.

PR #219, opened 2021-02-28
Add Capnp to Dockerfile

mishazharov:
This PR adds capnp to the Dockerfile because dependant PRs won't pass the CI before this is done


Phantomical (approved):
LGTM! Can you do a second PR to the README to add those to the install instructions?

PR #222, opened 2021-02-28
Add pkg-config to dependencies and update README

mishazharov:
(as in title)


Phantomical (approved):
LGTM with two things:
1. Please include why pkg-config is needed in the merge commit message.
2. Please also update the linux README instructions if `pkg-config` is also needed there.

PR #225, opened 2021-03-09
Add simple test to show usage for Capnp

mishazharov:
This is a small PR to demonstrate how Capnp works (and that we're linking with it properly etc.)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -51,7 +51,7 @@ list(APPEND sources ${CAPNP_SRCS})
 
 add_library(caffeine ${sources} ${headers})
 
-target_include_directories(caffeine PRIVATE "${CMAKE_CURRENT_BINARY_DIR}")
+target_include_directories(caffeine PUBLIC "${CMAKE_CURRENT_BINARY_DIR}")
 
 target_include_directories(caffeine
   PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}"
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -43,7 +43,7 @@ file(
 )
 
 
-set(CAPNPC_OUTPUT_DIR "${CMAKE_CURRENT_BINARY_DIR}/caffeine/")
+set(CAPNPC_OUTPUT_DIR "${CMAKE_CURRENT_BINARY_DIR}/gen/caffeine/")
 file(MAKE_DIRECTORY ${CAPNPC_OUTPUT_DIR})
 
 capnp_generate_cpp(CAPNP_SRCS CAPNP_HDRS "${capnp_schemas}")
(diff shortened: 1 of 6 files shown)


src/Protos/operation.capnp
```diff
@@ -1,5 +1,7 @@
 @0xf54292a58fc34e9a;
 
+using Cxx = import "/capnp/c++.capnp";
+$Cxx.namespace("caffeine_distributed::protos");
```
mishazharov, line 4: This places the operation file into the `caffeine_distributed::protos` namespace

Phantomical (approved):
Generally LGTM! 

As a suggestion: maybe designate a folder within the binary directory for generated artifacts. That way you don't have to worry too much about adding all sorts of weird directories in the include path. (Adding  `${CMAKE_BINARY_DIR}/test/unit` to the include path seems weird to me)

src/Protos/operation.capnp
Phantomical: Nit: Can we just do `caffeine::protos`? I would rather everything be in the `caffeine` root namespace.

src/CMakeLists.txt
```diff
@@ -43,15 +43,15 @@ file(
...
+set(CAPNPC_OUTPUT_DIR "${CMAKE_CURRENT_BINARY_DIR}/gen/caffeine/")
 file(MAKE_DIRECTORY ${CAPNPC_OUTPUT_DIR})
 
 capnp_generate_cpp(CAPNP_SRCS CAPNP_HDRS "${capnp_schemas}")
 list(APPEND sources ${CAPNP_SRCS})
 
 add_library(caffeine ${sources} ${headers})
 
-target_include_directories(caffeine PRIVATE "${CMAKE_CURRENT_BINARY_DIR}")
+target_include_directories(caffeine PUBLIC "${CMAKE_CURRENT_BINARY_DIR}/gen/")
```
Phantomical, line 54: I think you want `${CMAKE_BINARY_DIR}/gen` here?

test/unit/CMakeLists.txt
```diff
@@ -16,11 +16,12 @@ target_link_libraries(caffeine-unittest PRIVATE GTest::GTest)
 target_include_directories(caffeine-unittest PRIVATE "${CMAKE_SOURCE_DIR}/include")
 target_include_directories(caffeine-unittest PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}")
 target_include_directories(caffeine-unittest PRIVATE "${CMAKE_SOURCE_DIR}")
+target_include_directories(caffeine-unittest PRIVATE "${CMAKE_CURRENT_BINARY_DIR}")
```
Phantomical, line 19: I think the same applies here? Not sure what's supposed to be at that path.

PR #248, opened 2021-03-13
[13/N] Interpreter redesign - Move fcmp to ExprEvaluator

Phantomical:
As in title. Fixes a few edge cases I found that the previous version was not properly implementing.

/stack #246


test/run-pass/inst/fcmp/vector-fcmp-true.ll
```diff
@@ -0,0 +1,32 @@
+target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128"
+target triple = "x86_64-pc-linux-gnu"
+
+; Here we perform a vector fcmp true operation
+define dso_local void @test(float %x) local_unnamed_addr #0 {
```
Skrinikov, line 5: heh I definitely fully understand llvm code

Skrinikov (approved):

test/run-pass/inst/fcmp/vector-fcmp-true.ll
Phantomical: It has all the understandability of assembly with half the typing of java :P

PR #251, opened 2021-03-16
Add writable memory check

### LINKED ISSUE #180: Track whether memory is writable
Example test case that should fail
```c
#include "caffeine.h"

const char data[] = "abcdef";

__attribute__((noinline)) void write_to_data(char* mem) {
  *mem = 'u';
}

void test() {
  write_to_data((char*)data);
}
```

General instructions:
- Add a field to `Allocation` to indicating whether it's writable
- Modify `MemHeap::alloc` to to take a parameter that initializes whether the allocation is writable
- Fix up all call sites to now pass in the correct writability (there should be a LLVM attribute which indicates whether a newly created allocation is writable)
- Finally, make sure that writes to an allocation first check whether that allocation is writable and emit an error otherwise

Skrinikov:
Adds a writable memory field check to Allocation. 

NOTE: My CMAKE is messed up again so I am creating this PR to see whether the pipeline compiles

Fixes #180

### DIFF
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -1,3 +1,4 @@
 
 add_subdirectory(gen-test-main)
 add_subdirectory(gen-builtins)
+add_subdirectory(guided-fuzzing)
--- a/tools/guided-fuzzing/CMakeLists.txt
+++ b/tools/guided-fuzzing/CMakeLists.txt
@@ -1,0 +1,32 @@
+
+file(
+  GLOB sources
+  CONFIGURE_DEPENDS
+  RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}"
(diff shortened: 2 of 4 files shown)


include/caffeine/Memory/MemHeap.h
```diff
@@ -52,11 +52,13 @@ class Allocation {
 
   AllocationKind kind_;
 
+  bool is_writable;
```
Phantomical, line 55: As a suggestion: maybe make this an enum type? Its immediately obvious what something like `Permissions::Write` means whereas `true` isn't as clear. (Naming just a suggestion, feel free to use whatever names you  best)

Phantomical:
Also: I think you can probably fix the compile errors by rebasing/merging onto latest master.

PR #261, opened 2021-03-20
Change Executor to own Contexts

mishazharov:
In preparation for the multithreaded changes it's a good idea to make sure that the `Executor` owns the `Contexts` so that it can distribute work in any way it sees fit.


Phantomical (approved):
Awesome. LGTM.

A lot of the changes here look like formatting ones. This isn't a problem with your diff but I'm wondering why CI didn't catch them :/

mishazharov:
Yeah I noticed that but didn't realize that its pretty strange

PR #267, opened 2021-03-24
Adding thread safety

mishazharov:
This PR adds various mutex's in order to prepare for a multithreaded implementation of the Executor

### DIFF
--- a/include/caffeine/IR/Operation.h
+++ b/include/caffeine/IR/Operation.h
@@ -199,6 +199,7 @@ protected:
   uint16_t dummy_ = 0; // Unused, used for padding
 
   // Needs to be mutable so that const refs (ref<const Operation>) work.
+  // TODO: investigate if it's safe to use something besides std::memory_order_seq_cst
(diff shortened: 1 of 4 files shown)


include/caffeine/IR/Operation.h
```diff
@@ -194,10 +197,9 @@ class Operation : private CopyVTable {
 
   uint16_t opcode_;
   uint16_t dummy_ = 0; // Unused, used for padding
-  // When multithreading is implemented this will need to become atomic.
-  //
+
   // Needs to be mutable so that const refs (ref<const Operation>) work.
-  mutable uint32_t refcount = 0;
+  mutable std::atomic<uint32_t> refcount = 0;
```
Phantomical, line 202: This will use `std::memory_order_seq_cst` by default. Not something that needs to be solved in this PR but we should probably make some changes later on so that this is more efficient.

Phantomical (approved):
LGTM!

PR #277, opened 2021-03-28
Attempt to add multithreading

mishazharov:
This PR adds multithreading support to the `Executor`.


include/caffeine/Interpreter/Executor.h
```diff
@@ -3,16 +3,20 @@
 
 #include <cstdint>
 
+// Apparently semaphores only made it into c++20
+#include <boost/interprocess/sync/interprocess_semaphore.hpp>
```
Phantomical, line 7: I imagine an interprocess semaphore is going to be less efficient then a normal one. Maybe use [`boost::sync::semaphore`](https://www.boost.org/doc/libs/develop/libs/sync/doc/html/boost/sync/semaphore.html) instead?

src/Interpreter/Executor.cpp
```diff
@@ -27,10 +34,16 @@ bool Executor::has_next() const {
 
 void Executor::run() {
   while (has_next()) {
+    thread_finished.wait();
+
     auto ctx = next_context();
-    Interpreter interp(this, &ctx, logger);
+    auto thread = std::thread(run_single_context, this, std::move(ctx), logger,
```
Phantomical, line 40: Do we really need to spawn a thread for every context? Why not start off by spawning n threads and have each of them keep checking for contexts until there's a global signal to shut down?

src/Interpreter/PrintingFailureLogger.cpp
```diff
@@ -88,6 +88,7 @@ static void print_context_backtrace(std::ostream& os, const Context& ctx) {
 
 void PrintingFailureLogger::log_failure(const Model& model, const Context& ctx,
                                         const Failure& failure) {
+  std::unique_lock lock(mtx);
```
Phantomical, line 91: Why do we need the lock here? Could you print to a stringstream and then only hold the lock when writing the message to stdout.

src/Interpreter/Executor.cpp
```diff
@@ -27,10 +34,16 @@ bool Executor::has_next() const {
 
 void Executor::run() {
   while (has_next()) {
+    thread_finished.wait();
+
     auto ctx = next_context();
```
Phantomical, line 39: My guess as to the TSAN issue you're seeing here is that access to the queue isn't being synchronized here.

Phantomical (changes requested):
Here's my general suggestions:
- I'd recommend not using interprocess versions of semaphores for intra-process synchronization. The intra-process ones will be more efficient since they can take advantage of more OS funtionality.
- Please just create threads at startup and have them all basically run Executor::run. I'd suggest changing `next_context` to return a `std::optional<Context>` and to also be blocking. That way you can structure the main loop of `Executor::run` like this:
  ```cpp
  while (auto ctx = next_context()) {
    ...
  }
  ```
  I suspect that this would also fix your TSAN issue.

There's a few other bits around unnecessary locks that should be fixed.

include/caffeine/Interpreter/Executor.h
mishazharov: That primitive isn't actually in boost yet, but the other approach you suggested makes it possible to remove it.

src/Interpreter/Executor.cpp
```diff
@@ -2,16 +2,38 @@
...
 namespace caffeine {
 
+void run_worker(Executor* exec, FailureLogger* logger,
+                ExecutionContextStore* store) {
+  while (true) {
+    auto ctx = store->next_context();
+
+    if (!ctx.has_value()) {
+      return;
+    }
```
Phantomical, line 17: Nit: You can simplify this to
```suggestion
while (auto ctx = store->next_context()) {
```

src/bin/caffeine.cpp
```diff
@@ -162,8 +163,7 @@ int main(int argc, char** argv) {
 
   auto options = caffeine::ExecutorOptions{
       // TODO: Update once we actually support multithreaded execution
```
Phantomical, line 165: Remove the TODO?
```suggestion
```

src/Interpreter/Executor.cpp
```diff
@@ -2,16 +2,38 @@
 #include "caffeine/Interpreter/Interpreter.h"
 #include "caffeine/Interpreter/Store.h"
 
+#include <iostream>
```
Phantomical, line 5: `iostream` doesn't appear to be used?
```suggestion
```
We should really set up a lint for this :)

Phantomical (approved):
Some small comments but LGTM otherwise

README.md
```diff
@@ -10,6 +10,13 @@ The first build will take a while as it's going to build all the dependencies
 and install them within a `vcpkg_installed` directory within the repo folder.
 Once that's done you shouldn't have to worry about it anymore.
 
+## Debugging
+Running with TSAN can help detect a bunch of bugs in your code. To configure
+your build to run with TSAN do:
+```
+CC=clang-10 CXX=clang++-10 cmake -DCMAKE_C_FLAGS='-fsanitize=thread' -DCMAKE_CXX_FLAGS='-fsanitize=thread' ..
```
Phantomical, line 17: One last thing, maybe write this in terms of #303? (I realize this was added after you committed this)
```suggestion
CC=clang-10 CXX=clang++-10 cmake -DCAFFEINE_ENABLE_TSAN=ON ..
```

PR #294, opened 2021-04-03
Introduce tool to generate implementations of LLVM builtins

Phantomical:
The current state of the repo is that we have builtin implementations for the most common calls to the `llvm.memset.*` intrinsic. Namely, we really only support the 
```
void llvm.memset.p0i8.i64(i8* %dst, i8 %val, i64 %len, i1 %isvolatile)
```
variant. This is done by hand-implementing a function
```
void caffeine_builtin_memset(...)
```
with the same signature in _LLVM IR_ and then having the interpreter patch calls to the intrinsic to instead point to the function we implemented by hand.

This has some big limitations:
1. We don't support anything other than the most basic memset calls. Anything that gets generated to use a different type will fail to compile. As an example, this will fail on platforms where pointers are 32-bits in size.
2. We don't support any platform that is not data-layout-compatible with x86_64-linux-gnu. This is the more relevant issue because it means that we cannot compile our test cases for windows.

Both limitations are decently large problems. The solution I've undertaken here is to instead generate the required builtin
implementations at build time. This means that we never end up needing an unsupported function variant since we support _all_ the variants.

As a bonus, we can now also use IR names that can't be defined from C.

This PR is part of my work to address #293

### DIFF
--- a/tools/gen-builtins/main.cpp
+++ b/tools/gen-builtins/main.cpp
@@ -52,7 +52,6 @@ int main(int argc, char** argv) {
   // parse for LLVM 10.0.0. To work around that I'm printing using textual IR.
   if (OutputFilename == "-") {
     module->print(llvm::outs(), nullptr);
-    // WriteBitcodeToFile(*m, llvm::outs());
   } else {
     std::error_code EC;
     llvm::raw_fd_ostream os(OutputFilename, EC);
@@ -65,6 +64,5 @@ int main(int argc, char** argv) {
     }
 
     module->print(os, nullptr);
-    // WriteBitcodeToFile(*m, os);
   }
 }
--- a/tools/gen-builtins/main.cpp
+++ b/tools/gen-builtins/main.cpp
@@ -1,0 +1,70 @@
+
+#include "builtins.h"
+#include "caffeine/Support/Assert.h"
+#include <fmt/format.h>
+#include <llvm/IR/LLVMContext.h>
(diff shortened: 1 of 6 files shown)


tools/gen-builtins/main.cpp
```diff
@@ -0,0 +1,70 @@
...
+      caffeine::generateMemset(module.get(), &function);
+      continue;
+    }
+  }
+
+  // NOTE: WriteBitcodeToFile seems to produce something that llvm-dis can't
+  // parse for LLVM 10.0.0. To work around that I'm printing using textual IR.
+  if (OutputFilename == "-") {
+    module->print(llvm::outs(), nullptr);
+    // WriteBitcodeToFile(*m, llvm::outs());
```
mishazharov, line 55: Delete?

tools/gen-builtins/main.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+    if (EC) {
+      errs() << argv[0] << ": ";
+      WithColor::error() << " unable to write to output file '"
+                         << OutputFilename << "': " << EC.message() << "\n";
+      return 1;
+    }
+
+    module->print(os, nullptr);
+    // WriteBitcodeToFile(*m, os);
```
mishazharov, line 68: delete?

mishazharov (approved):
Is this new code actually being used? Was expecting the old memset and resolve to be stripped out

PR #299, opened 2021-04-03
Break `Solver` out of `Context`

mishazharov:
This PR breaks out the `Solver` from the `Context`. This is helpful in order to allow multiple `Solver`s to coexist in a multithreaded environment. As part of this change, the `Interpreter` now creates the `Solver` stack, as opposed to the caffeine binary harness. This is to allow each `Interpreter` to own it's `Solver` so that we can have a `Solver` per thread. Previously, it was only possible to have one `Solver` per runtime.

I was considering adding a method to `Context` along the lines of `setSolver` but decided against it because of the ownership issue.

### DIFF
--- a/src/Interpreter/Interpreter.cpp
+++ b/src/Interpreter/Interpreter.cpp
@@ -4,6 +4,10 @@
 #include "caffeine/Interpreter/StackFrame.h"
 #include "caffeine/Interpreter/Store.h"
 #include "caffeine/Interpreter/Value.h"
+#include "caffeine/Solver/CanonicalizingSolver.h"
+#include "caffeine/Solver/SequenceSolver.h"
+#include "caffeine/Solver/SimplifyingSolver.h"
+#include "caffeine/Solver/Z3Solver.h"
 #include "caffeine/Support/Assert.h"
 
 #include <boost/range/adaptor/transformed.hpp>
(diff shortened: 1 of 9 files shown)


src/Interpreter/Interpreter.cpp
```diff
@@ -26,11 +31,14 @@ Interpreter::Interpreter(Context* ctx, ExecutionPolicy* policy,
                          ExecutionContextStore* store, FailureLogger* logger,
                          const InterpreterOptions& options)
     : policy(policy), store(store), ctx(ctx), logger(logger), options(options) {
+  solver = caffeine::make_sequence_solver(caffeine::SimplifyingSolver(),
```
Phantomical, line 34: This will work for now but we're going to want the actual solver to be both configurable and sharable as much as possible (modulo needing 1 solver per thread) in the future.

Phantomical (approved):
LGTM. I don't particularly like that `Interpreter` now creates a solver stack but it will work for now.

Can you add an explanation of why this is necessary to both the commit message and the PR comment? It took me a while to figure out why so having this spelled out would be helpful if we're looking back at it.

src/Interpreter/Interpreter.cpp
mishazharov: Yep, I agree. We can change it when we need to.

PR #301, opened 2021-04-03
Add support for `llvm.memcpy` to gen-builtins

Phantomical:
This extends the support for generating builtins in #294 to also support `llvm.memcpy`. See the comment within the file for details on what actually gets generated.

Note that we can't delete `caffeine_builtin_memcpy` yet since it's still used by `memmove`.

/stack #297 

This is part of the work towards addressing #293


mishazharov (approved):

mishazharov:
Nice refactors!

PR #307, opened 2021-04-06
Reimplement `PersistentArray` using the immer library

Phantomical:
I looked through the implementation of `PersistentArray`, spotted some problems, and decided to shell out to a dedicated library for these kind of things. `immer` is a library offering a number of persistent data structures. Now that we can can more easily add downloaded dependencies due to #305, I felt that it was worthwhile to use a well-tested library here. As a bonus, `immer::vector` is also thread-safe (with the default setup) so we can avoid having to wrap everything in `std::shared_mutex`.

This PR does three things:
1. Add `immer` as a downloadable project. I also had to add some scripts because `ExternalProject_Add`'s patch step occasionally will get reapplied to an already patched version of the code base.
2. Rewrite `PersistentArray` to be a thin wrapper around `immer::vector`.
3. Cleanup one instance of code where it could be done more efficiently now.

/stack #305


Skrinikov (approved):
Glanced over it. Looks mostly just cmake stuff and removing persistent array. LEt me know if you need a more thorough review

Phantomical:
No that's about it. It doesn't entirely remove `PersistentArray` (although the github UI makes it look like it does) but I think we're good here.

PR #324, opened 2021-04-11
Implement new `SparseVector` data structure

Phantomical:
This is vector which can have holes where elements might be. This allows us to erase elements while iterating over the same vector which is useful for transforms that will be applied to the list of assertions.

/stack #323

### DIFF
--- a/test/unit/ADT/SparseVector.cpp
+++ b/test/unit/ADT/SparseVector.cpp
@@ -1,0 +1,80 @@
+#include "caffeine/ADT/SparseVector.h"
+#include <gtest/gtest.h>
+
+using namespace caffeine;
+
+TEST(SparseVectorTests, eraseExcludedFromIterationOrder) {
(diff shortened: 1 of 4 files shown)


test/unit/ADT/SparseVector.cpp
```diff
@@ -0,0 +1,80 @@
+#include "caffeine/ADT/SparseVector.h"
+#include <gtest/gtest.h>
+
+using namespace caffeine;
+
+TEST(SparseVectorTests, eraseExcludedFromIterationOrder) {
```
mishazharov, line 6: What happens when you have an iterator to an element that's deleted?

mishazharov (approved):

test/unit/ADT/SparseVector.cpp
Phantomical: It's no longer valid to derefence that iterator but you can still increment it to get to the next element (after the slot that would have been there).

PR #331, opened 2021-04-13
Add transparent interning for operations

Phantomical:
This PR adds code to transparently intern operation references at creation time. It does this by maintaining a global map containing `weak_ptr`s to all `Operation` instances. This allows it to transparently reuse operation instances while they are being created.

This can then be reused by downstream systems to efficiently deduplicate expressions.


brookedolny (approved):
Looks good

PR #333, opened 2021-04-13
Take advantage of `AssertionList` features

### LINKED ISSUE #320: Refactor the list of assertions to better keep track of what is known to be true
Right now we just store assertions in a `std::vector`. 

By doing this we should be able to get the following benefits:
- We can now incrementally simplify the whole set of assertions
- Solvers can now perform some of query splitting techniques that show up in the literature

Phantomical:
This takes advantage of the features of `AssertionList` to avoid performing solver queries if not necessary.

Closes #320


mishazharov (approved):
LGTM!

PR #334, opened 2021-04-14
Add GuidedExecutionPolicy

mishazharov:
This PR adds a `GuidedExecutionPolicy` class. The purpose of this class will be to load it with assertions in order to only allow execution of paths that fit a given concrete input to a function.

### DIFF
--- a/tools/guided-fuzzing/include/GuidedExecutionPolicy.h
+++ b/tools/guided-fuzzing/include/GuidedExecutionPolicy.h
@@ -1,0 +1,42 @@
+#pragma once
+
+#include "caffeine/Interpreter/Context.h"
+#include "caffeine/Interpreter/Policy.h"
+#include "caffeine/Solver/Solver.h"
+
+namespace caffeine {
+
+/**
+ * Class to control the behaviour of an Interpreter instance. It also provides a
+ * number of hooks for the policy to execute code when various context-related
+ * events occur within the executor. By default, these do nothing.
+ *
+ * If you don't care about controlling queuing behaviour you may want to derive
+ * from AlwaysAllowExecutionPolicy instead of the base ExecutionPolicy class.
+ */
+class GuidedExecutionPolicy : ExecutionPolicy {
+  AssertionList requiredAssertions_;
+  std::shared_ptr<Solver> solver_;
+
+public:
+  GuidedExecutionPolicy(AssertionList& list, std::shared_ptr<Solver> solver);
+  ~GuidedExecutionPolicy() = default;
+
+  /**
+   * Called when a context forks to determine whether we should continue
+   * executing that branch.
+   *
+   * Note that this method does not control whether dead branches continue to
+   * execute. Those will be pruned irregardless of what this method returns.
+   */
+  virtual bool should_queue_path(const Context& ctx);
+
+protected:
+  GuidedExecutionPolicy(GuidedExecutionPolicy&&) = default;
+  GuidedExecutionPolicy(const GuidedExecutionPolicy&) = default;
+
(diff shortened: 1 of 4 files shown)


tools/guided-fuzzing/src/GuidedExecutionPolicy.cpp
```diff
@@ -0,0 +1,26 @@
...
+  if (solver_->check(list) == UNSAT) {
+    CAFFEINE_ABORT("GuidedExecutionPolicy requires satisfiable premise");
+  }
+}
+
+bool GuidedExecutionPolicy::should_queue_path(const Context& ctx) {
+  AssertionList combined;
+  for (const auto& i : ctx.assertions) {
+    combined.insert(i);
+  }
```
Phantomical, line 17: Just copying the list keeps the internal state better and is probably faster.
```suggestion
  AssertionList combined = ctx.assertions;
```

tools/guided-fuzzing/include/GuidedExecutionPolicy.h
```diff
@@ -0,0 +1,42 @@
...
+namespace caffeine {
+
+/**
+ * Class to control the behaviour of an Interpreter instance. It also provides a
+ * number of hooks for the policy to execute code when various context-related
+ * events occur within the executor. By default, these do nothing.
+ *
+ * If you don't care about controlling queuing behaviour you may want to derive
+ * from AlwaysAllowExecutionPolicy instead of the base ExecutionPolicy class.
+ */
```
Phantomical, line 16: Can you update this to a comment which actually describes the `GuidedExecutionPolicy`? Copying the comment from the base class really doesn't tell us anything new.

tools/guided-fuzzing/include/GuidedExecutionPolicy.h
```diff
@@ -0,0 +1,42 @@
...
+  GuidedExecutionPolicy(AssertionList& list, std::shared_ptr<Solver> solver);
+  ~GuidedExecutionPolicy() = default;
+
+  /**
+   * Called when a context forks to determine whether we should continue
+   * executing that branch.
+   *
+   * Note that this method does not control whether dead branches continue to
+   * execute. Those will be pruned irregardless of what this method returns.
+   */
```
Phantomical, line 31: Same thing here. Either update the comment to indicate what this method does or just delete it.

tools/guided-fuzzing/include/GuidedExecutionPolicy.h
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Class to control the behaviour of an Interpreter instance. It also provides a
+ * number of hooks for the policy to execute code when various context-related
+ * events occur within the executor. By default, these do nothing.
+ *
+ * If you don't care about controlling queuing behaviour you may want to derive
+ * from AlwaysAllowExecutionPolicy instead of the base ExecutionPolicy class.
+ */
+class GuidedExecutionPolicy : ExecutionPolicy {
```
Phantomical, line 17: Needs to be public otherwise this class won't be usable as an `ExecutionPolicy`
```suggestion
class GuidedExecutionPolicy : public ExecutionPolicy {
```

Phantomical (approved):
I've left some comments. Approved conditional on them being fixed.

Not something you have to address, but outside of the main libcaffeine we probably don't need to split things into separate `include` and `src` directories.

tools/guided-fuzzing/include/GuidedExecutionPolicy.h
mishazharov: Nice catch

PR #336, opened 2021-04-15
[180] Add writable memory permissions and checks

### LINKED ISSUE #180: Track whether memory is writable

Skrinikov:
As per comments from last PR I changed the boolean flag to an general enum type

Fixes #180

### DIFF
--- a/include/caffeine/Memory/MemHeap.h
+++ b/include/caffeine/Memory/MemHeap.h
@@ -35,7 +35,12 @@ enum class AllocationKind { Alloca, Malloc, Global };
  * with ReadWrite permissions a user could can read and modify the underlying
  * memory. For a Read, a user can read the data, but cannot modify the buffer.
  */
-enum class AllocationPermissions { None, Read, Write, ReadWrite };
+enum class AllocationPermissions { 
+  None = 0, 
+  Read = 1, 
+  Write = 2, 
+  ReadWrite = 3,
+};
 
 /**
  * A memory allocation (either alive or dead).
--- a/include/caffeine/Memory/MemHeap.h
+++ b/include/caffeine/Memory/MemHeap.h
@@ -35,7 +35,7 @@ enum class AllocationKind { Alloca, Malloc, Global };
  * with ReadWrite permissions a user could can read and modify the underlying
  * memory. For a Read, a user can read the data, but cannot modify the buffer.
  */
-enum class AllocationPermissions { None, Read, Write, ReadWrite};
+enum class AllocationPermissions { None, Read, Write, ReadWrite };
 
 /**
  * A memory allocation (either alive or dead).
@@ -214,7 +214,8 @@ public:
    * This will add the corresponding assertions to the context as well.
    */
   AllocId allocate(const OpRef& size, const OpRef& alignment, const OpRef& data,
-                   AllocationKind kind, AllocationPermissions permissions, Context& ctx);
+                   AllocationKind kind, AllocationPermissions permissions,
+                   Context& ctx);
 
   /**
    * Deallocate an existing allocation.
--- a/include/caffeine/Memory/MemHeap.h
+++ b/include/caffeine/Memory/MemHeap.h
@@ -35,10 +35,10 @@ enum class AllocationKind { Alloca, Malloc, Global };
  * with ReadWrite permissions a user could can read and modify the underlying
  * memory. For a Read, a user can read the data, but cannot modify the buffer.
  */
-enum class AllocationPermissions { 
-  None = 0, 
-  Read = 1, 
-  Write = 2, 
+enum class AllocationPermissions {
+  None = 0,
+  Read = 1,
+  Write = 2,
   ReadWrite = 3,
 };
 
--- a/include/caffeine/Memory/MemHeap.h
+++ b/include/caffeine/Memory/MemHeap.h
@@ -28,6 +28,16 @@ class Solver;
 enum class AllocationKind { Alloca, Malloc, Global };
 
 /**
+ * Permisisons for an allocation
+ *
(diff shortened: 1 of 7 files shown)


include/caffeine/Memory/MemHeap.h
```diff
@@ -27,6 +27,16 @@ class Solver;
...
 
+/**
+ * Permisisons for an allocation
+ *
+ * This is used to indicate what kind of operations a user is allowed
+ * to perfom on the requested allocation. As an example, for an allocation
+ * with ReadWrite permissions a user could can read and modify the underlying
+ * memory. For a Read, a user can read the data, but cannot modify the buffer.
+ */
+enum class AllocationPermissions { None, Read, Write, ReadWrite };
```
Phantomical, line 38: If we make this a bitfield then you can easily do `is_readable` or `is_writable checks.

```suggestion
enum class AllocationPermissions { 
  None = 0,
  Read = 1,
  Write = 2, 
  ReadWrite = 3
};
```

I do realize this doesn't change the semantics at all but it makes the values explicit so that they can be relied upon :)

src/Memory/MemHeap.cpp
```diff
@@ -16,20 +16,29 @@ namespace caffeine {
...
 Allocation::Allocation(const OpRef& address, const ConstantInt& size,
-                       const OpRef& data, AllocationKind kind)
-    : Allocation(address, make_ref<Operation>(size), data, kind) {}
+                       const OpRef& data, AllocationKind kind,
+                       AllocationPermissions permissions)
+    : Allocation(address, make_ref<Operation>(size), data, kind, permissions) {}
 
 void Allocation::overwrite(const OpRef& newdata) {
+  CAFFEINE_ASSERT(perms_ == AllocationPermissions::Write ||
+                      perms_ == AllocationPermissions::ReadWrite,
```
Phantomical, line 34: Using the bitfield idea:
```suggestion
  CAFFEINE_ASSERT(perms_ & AllocationPermissions::Write,
```

src/Memory/MemHeap.cpp
```diff
@@ -16,20 +16,29 @@ namespace caffeine {
...
+                  "tried to write to unwritable allocation");
   data_ = newdata;
 }
 void Allocation::overwrite(OpRef&& newdata) {
+  CAFFEINE_ASSERT(perms_ == AllocationPermissions::Write ||
+                      perms_ == AllocationPermissions::ReadWrite,
```
Phantomical, line 40: Same here:
```suggestion
  CAFFEINE_ASSERT(perms_ & AllocationPermissions::Write,
```

src/Memory/MemHeap.cpp
```diff
@@ -166,6 +175,9 @@ void Allocation::write(const OpRef& offset, const OpRef& value_,
 
   CAFFEINE_ASSERT(offset->type().is_int(),
                   "tried to write at non-integer offset");
+  CAFFEINE_ASSERT(perms_ == AllocationPermissions::Write ||
+                      perms_ == AllocationPermissions::ReadWrite,
```
Phantomical, line 179:
```suggestion
  CAFFEINE_ASSERT(perms_ & AllocationPermissions::Write,
```

src/Memory/MemHeap.cpp
```diff
@@ -204,6 +216,9 @@ void Allocation::write(const OpRef& offset, const LLVMScalar& value,
 void Allocation::write(const OpRef& offset, llvm::Type* type,
                        const LLVMValue& value, const MemHeap& heap,
                        const llvm::DataLayout& layout) {
+  CAFFEINE_ASSERT(perms_ == AllocationPermissions::Write ||
+                      perms_ == AllocationPermissions::ReadWrite,
```
Phantomical, line 220:
```suggestion
  CAFFEINE_ASSERT(perms_ & AllocationPermissions::Write,
```

src/Interpreter/ExprEval.cpp
```diff
@@ -356,7 +356,7 @@ OpRef ExprEvaluator::visitGlobalData(llvm::Constant& constant, unsigned AS) {
       bitwidth, layout.getTypeAllocSizeInBits(type).getFixedSize()));
   Allocation alloc{ConstantInt::CreateZero(bitwidth), size,
                    AllocOp::Create(size, ConstantInt::CreateZero(8)),
-                   AllocationKind::Alloca};
+                   AllocationKind::Alloca, AllocationPermissions::ReadWrite};
   alloc.write(ConstantInt::CreateZero(bitwidth), type, value, ctx->heap,
               layout);
```
Phantomical, line 361: At the end of this you probably want to update the permissions to be either `Read` or `ReadWrite` depending on what attributes the global has. I'm not sure which attributes are relevant here. You'll have to figure that out from the llvm reference page.

Phantomical (commented):
This approach looks good to me! Currently everything is set as `ReadWrite` so I think you might want to try and get it so that the test case in the linked issue passes.

If you want to do that in a second PR then that also works (but then this one shouldn't be closing that issue). If this is the case then LMK and I'll approve once the other comments are addressed :)

Phantomical (approved):
LGTM!

Issue #335, opened 2021-04-15
Optimizations tracking issue


Phantomical:
Currently our interpreter is quite slow and we'd like it to be faster. This issue is meant to track various performance optimization PRs.

PR #341, opened 2021-04-16
Elide Z3 solver calls if the result is already known

Phantomical:
As in title. This also avoids creating `z3::tactic` instances unnecessarily as that is quite expensive.

This is part #335

### DIFF
--- a/src/Solver/Z3Solver.cpp
+++ b/src/Solver/Z3Solver.cpp
@@ -209,6 +209,11 @@ Z3Solver::~Z3Solver() {}
 
 SolverResult Z3Solver::check(AssertionList& assertions,
                              const Assertion& extra) {
+  if (assertions.unproven().empty() && extra.is_constant_value(true))
+    return SolverResult::SAT;
+  if (extra.is_constant_value(false))
+    return SolverResult::UNSAT;
+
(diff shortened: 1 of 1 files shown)


src/Solver/Z3Solver.cpp
```diff
@@ -209,6 +209,11 @@ Z3Solver::~Z3Solver() {}
 
 SolverResult Z3Solver::check(AssertionList& assertions,
                              const Assertion& extra) {
+  if (assertions.unproven().empty() && extra.is_constant_value(true))
+    return SolverResult::SAT;
+  if (extra.is_constant_value(false))
+    return SolverResult::UNSAT;
```
mishazharov, line 215: Do you need `assertions.unproven().empty()` on the second `if` too?

mishazharov (approved):

src/Solver/Z3Solver.cpp
Phantomical: No. If the extra assertion is false then it doesn't matter what the rest of assertions are (it's logically an and)

mishazharov: True, totally missed that

PR #346, opened 2021-04-16
Update to LLVM 11

mishazharov:
Using LLVM 11 with AFL++ provides good performance gains, and since it's
available in Ubuntu we can just upgrade it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,9 +9,9 @@ ARG DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update \
     && apt-get -y install \
-        llvm-10-dev \
-        clang-10 \
-        clang++-10 \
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -28,3 +28,4 @@ RUN apt-get update \
         capnproto \
         pkg-config \
     && rm -rf /var/lib/apt/lists/*
+RUN sudo update-alternatives --install /usr/local/bin/llvm-config llvm-config /usr/bin/llvm-config-11 20
```
Phantomical, line 31: Can you add this to the previous command? That way we don't create any extra layers

Phantomical (approved):
LGTM. We should try and figure out what LLVM versions we want to support rather than just picking one in the future though.

PR #351, opened 2021-04-17
Convert IR transformation modules into a LLVM opt pass plugin

Phantomical:
This PR takes the `gen-builtins` and `gen-test-main` binaries and converts them into a LLVM opt pass plugin. This removes a bunch of code around handling arguments and parsing input files.

### Motivation
- Having them all in one place means fewer cmake commands to juggle to get tests setup
- It'll scale better when we need to add more transformations 
- Opt can properly output bitcode (unlike `llvm::WriteBitcodeToFile` for some reason) so it works with the rest of the compilation pipeline
- We no longer need to have a copy of reading/writing bitcode for each optimization we want to make

Extra note: This currently disables ASAN in CI as it really doesn't like being used within plugins. I'll create a follow-up PR to reenable ASAN in CI.


mishazharov (approved):
Noice!

PR #354, opened 2021-04-17
Implement backtrace printer for Context

Phantomical:
This implements a detailed backtrace printer for printing the call stack of the program being executed. If the information is available it will try to print as much context info as possible. However, I've tried to write it to still work acceptably and not crash even if the context is quite broken as this is intended to be used within a custom assertion type I'll be introducing in the next few PRs.

An example backtrace  is:
```
 #0 test
     | store i32 0, i32* %2, align 4, !tbaa !2
 #1 main
     | call void @test()
```

### DIFF
--- a/include/caffeine/Interpreter/Context.h
+++ b/include/caffeine/Interpreter/Context.h
@@ -11,6 +11,7 @@
 #include <llvm/IR/Constant.h>
 #include <llvm/IR/Function.h>
 
+#include <iosfwd>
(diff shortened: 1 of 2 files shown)


Phantomical:
The build failure seems to be unrelated? I'm going to wait until #351 and #352 land since those will probably address the issue.

mishazharov (approved):

PR #356, opened 2021-04-17
Add support for recovering from an unimplemented feature

Phantomical:
This PR allows for caffeine to recover from running into an unsupported LLVM operation by killing the currently executing context. To do this it introduces two new assertion/abort macros:
- `CAFFEINE_UNSUPPORTED(message)`, and
- `CAFFEINE_UASSERT(cond, message)`

These behave like `CAFFEINE_ABORT` and `CAFFEINE_ASSERT`, respectively, but instead of aborting the whole process will instead (depending on configuration) throw an exception that gets caught within executor. This means that we end up killing the context but not the whole process, allowing other contexts to continue executing. Since we know that things aren't entirely broken we also have the ability to do things like print out a nice call stack of the current context.

/stack #354

### DIFF
--- a/src/Interpreter/ExprEval.cpp
+++ b/src/Interpreter/ExprEval.cpp
@@ -4,6 +4,7 @@
 #include "caffeine/Interpreter/Value.h"
 #include "caffeine/Memory/MemHeap.h"
 #include "caffeine/Support/LLVMFmt.h"
+#include "caffeine/Support/UnsupportedOperation.h"
 #include <fmt/format.h>
 #include <fmt/ostream.h>
 #include <llvm/IR/Constants.h>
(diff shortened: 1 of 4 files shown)


src/Interpreter/ExprEval.cpp
```diff
@@ -92,10 +93,9 @@ LLVMValue ExprEvaluator::evaluate(llvm::Value* val) {
   if (auto* cnst = llvm::dyn_cast<llvm::GlobalVariable>(val))
     return visitGlobalVariable(*cnst);
 
-  std::cout << magic_enum::enum_name((llvm::Value::ValueTy)val->getValueID())
-            << std::endl;
-
-  CAFFEINE_ABORT(fmt::format("Unsupported expression: {}", *val));
+  CAFFEINE_UNSUPPORTED(fmt::format(
```
mishazharov, line 96: I think we also have a bunch of asserts related to resizable vectors?

mishazharov (approved):
noice

src/Interpreter/ExprEval.cpp
Phantomical: Yeah I'll do a follow-up PR to go through and fix all of these.

PR #357, opened 2021-04-18
Add AFLplusplus for symbolic mutator

mishazharov:
This PR fixes a bug where empty strings cannot be passed to the `caffeine_dependency` rule and also adds AFL++ to the symbolic mutator / guided fuzzing tool.

### DIFF
--- a/cmake/AFLplusplus.patch
+++ b/cmake/AFLplusplus.patch
@@ -1,0 +1,19 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+new file mode 100644
+index 00000000..a5991245
+--- /dev/null
++++ b/CMakeLists.txt
+@@ -0,0 +1,13 @@
++cmake_minimum_required(VERSION 3.8)
++
++project(AFLplusplus VERSION "3.12" LANGUAGES C)
++include(CMakePackageConfigHelpers)
++add_library(${PROJECT_NAME} INTERFACE)
++target_include_directories(${PROJECT_NAME}
++  INTERFACE
++  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
++  $<INSTALL_INTERFACE:include/AFLplusplus>
++)
++install(TARGETS ${PROJECT_NAME} EXPORT ${PROJECT_NAME}Config)
++install(EXPORT ${PROJECT_NAME}Config DESTINATION lib/cmake/${PROJECT_NAME})
++install(DIRECTORY include/ DESTINATION include/${PROJECT_NAME})
--- a/cmake/AFLplusplus.patch
+++ b/cmake/AFLplusplus.patch
@@ -3,7 +3,10 @@ new file mode 100644
 index 00000000..a5991245
 --- /dev/null
 +++ b/CMakeLists.txt
-@@ -0,0 +1,13 @@
+@@ -0,0 +1,16 @@
++# This doesn't actually need any build steps because we just care about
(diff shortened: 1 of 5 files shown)


cmake/CaffeineDependencies.cmake
```diff
@@ -35,7 +35,7 @@ function(caffeine_dependency PACKAGE VERSION)
   make_directory("${SUB_DIR}")
 
   set(QUOTED_ARGS "")
-  foreach(ARG ${ARGN})
+  foreach(ARG IN LISTS ARGN)
```
Phantomical, line 38: I though this was supposed to work in the first place 😠 Goddamn cmake. Will use that next time.

tools/guided-fuzzing/CMakeLists.txt
```diff
@@ -26,7 +26,22 @@ target_include_directories(afl-guided-mutator SYSTEM
   PRIVATE "${FMT_INCLUDE_DIRS}"
 )
 
+if (CMAKE_VERSION VERSION_GREATER "3.16")
+  cmake_policy(SET CMP0097 NEW)
+endif()
```
Phantomical, line 31: Maybe put this in the top-level CMakeLists? It's a global change so that seems like the right place for it

tools/guided-fuzzing/CMakeLists.txt
```diff
@@ -26,7 +26,22 @@ target_include_directories(afl-guided-mutator SYSTEM
...
+caffeine_dependency(
+  AFLplusplus                ""
+  GIT_REPOSITORY             https://github.com/AFLplusplus/AFLplusplus
+  GIT_TAG                    3.12c
+  GIT_SHALLOW                TRUE
+  GIT_SUBMODULES             ""
+  PATCH_COMMAND  "${CMAKE_COMMAND}"
+    "${CMAKE_SOURCE_DIR}/cmake/AFLplusplus.patch"
+    -P "${CMAKE_SOURCE_DIR}/cmake/CaffeinePatch.cmake"
+)
```
Phantomical, line 42: Can you put this in `CaffeineDependencies.cmake`? I'd rather have all our dependencies in one place.

cmake/AFLplusplus.patch
```diff
@@ -0,0 +1,19 @@
...
++include(CMakePackageConfigHelpers)
++add_library(${PROJECT_NAME} INTERFACE)
++target_include_directories(${PROJECT_NAME}
++  INTERFACE
++  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
++  $<INSTALL_INTERFACE:include/AFLplusplus>
++)
++install(TARGETS ${PROJECT_NAME} EXPORT ${PROJECT_NAME}Config)
++install(EXPORT ${PROJECT_NAME}Config DESTINATION lib/cmake/${PROJECT_NAME})
++install(DIRECTORY include/ DESTINATION include/${PROJECT_NAME})
```
Phantomical, line 19: Question: Does building AFL not involve actually compiling anything?

Phantomical (approved):
LGTM with comments resolved.

cmake/CaffeineDependencies.cmake
mishazharov: ikr!

cmake/AFLplusplus.patch
mishazharov: I'll add a comment about why we don't need to build it

PR #364, opened 2021-04-22
Implement support for debug intrinsics

Phantomical:
There isn't anything special we need to do to support this, it's just a matter of not crashing when we encounter any of them. That is all this PR does.

To help with the better stack traces that are being introduced, and to test that this works, this PR compiles all of the IR with debug info.


Phantomical:
Turns out this was already done in #358

PR #365, opened 2021-04-23
Fix zext bug within `Z3Solver`

### LINKED ISSUE #361: `Z3Solver` crashes when trying to `zext` a boolean type
The summary of this issue is that the solver calls `operand.sort().bv_size()` even when the operand is not a bitvector (this happens in the case of booleans). Z3, naturally, doesn't like this and throws an exception.

Test case:
```#include "caffeine.h"
#include <stdint.h>
#include <stdbool.h>
#include <stddef.h>

typedef unsigned (*func_t)(unsigned);

unsigned func1(unsigned x) {
  return x * 20;
}

__attribute__((optnone)) func_t getfunc(bool x) {
  return x ? func1 : NULL;
}

void test(uint8_t x, uint32_t y) {
  func_t func = getfunc(x);

  caffeine_assert(func(y) == y * 20);
}
```
Unfortunately this test case is rather tied up with some of my function-pointer related changes and not something that will work on current master. I will try and create a better test case tomorrow.

Phantomical:
This PR fixes a bug where caffeine will crash when `Z3Solver` attempts to build a `zext` operation for an inner type which is actually represented using a boolean. See the commit messages for a more detailed description of the bug. I've included a test case to ensure that the issue doesn't crop up again.

Closes #361

### DIFF
--- a/src/Solver/Z3Solver.cpp
+++ b/src/Solver/Z3Solver.cpp
@@ -591,13 +591,13 @@ z3::expr Z3OpVisitor::visitZExt(const UnaryOp& op) {
   auto operand = visit(*op.operand());
   auto src = normalize_to_bv(operand);
 
-  return z3::zext(src, op.type().bitwidth() - operand.get_sort().bv_size());
+  return z3::zext(src, op.type().bitwidth() - src.get_sort().bv_size());
 }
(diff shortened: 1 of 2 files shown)


mishazharov (approved):
Damn, that was a silly mistake on my part. Nice catch

PR #376, opened 2021-04-29
Simplify evaluation of zeroinitializer globals

Phantomical:
In this case we already know exactly what value the global will have so there's no need to go an evaluate each sub-expression since we know the final result will just be zeroed memory.

This also fixes an issue where allocations would be 8 times larger than they needed to be (due to using size in bits instead of size in bytes) as well as an issue where the pass to generate `main` for test cases was not nul-terminating strings.


mishazharov (approved):
Noice

PR #380, opened 2021-05-07
Add some simple benchmarks 

### LINKED ISSUE #203: Create benchmark suite
This should let us check whether we are regressing or improving performance.

Phantomical:
This PR adds a `bench` folder along with two benchmarks lifted out of a KLEE tutorial. The two benchmarks added right now are
1. The maze example from the KLEE tutorial. This benchmark is composed mainly of concrete variables with the exception of the input. Currently this takes 1m 23s of total CPU time.
2. A symbolic version of the maze example. This benchmark makes everything symbolic. This benchmark currently takes > 1h 45m of total CPU time (I didn't wait to see how long it would take)

If you read the examples closely you'll note that there's a bug in the maze algorithm. This means that there are multiple possible paths through the maze. I've also cleaned up the internal cmake setup somewhat.

This is part of #335 and closes #203. Feel free to suggest more benchmarks you think we should add :)

### DIFF
--- a/bench/maze.c
+++ b/bench/maze.c
@@ -1,6 +1,11 @@
 
 #include "caffeine.h"
 
+/**
+ * This benchmark is a direct translation of klee's maze example. It has a bunch
+ * of branches but all values end up being concrete on any of those branches.
+ */
+
 #define H 7
 #define W 11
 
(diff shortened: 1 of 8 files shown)


bench/maze.c
```diff
@@ -0,0 +1,79 @@
+
+#include "caffeine.h"
+
+/**
+ * This benchmark is a direct translation of klee's maze example. It has a bunch
+ * of branches but all values end up being concrete on any of those branches.
+ */
```
mishazharov, line 7: Can you link it please?

mishazharov (approved):
LGTM

PR #381, opened 2021-05-08
Optimize constant-size symbolic arrays

Phantomical:
Currently, we use a symbolic array to represent the arrays emitted `caffeine_make_symbolic`. This is quite slow. This PR replaces this with allocating a `FixedArray` of byte constants. To make this work, I've changed `Context` to have a map of symbol names to constant values.

## Performance
This change improves the peformance on the `maze-symbolic` benchmark by at least 100x. I haven't actually had the time to run that benchmark to completion on the old version of the code so it might be more.

## Detailed Changes
- Change `Model::evaluate` to automatically decide the values of constants that are not mentioned in the model (these are ones where we don't care about their value).
- Change `PrintingFailureLogger` to directly use the new way of exposing named constants.
- Add a map of string constant names -> expressions for their values to `Context`.
- Change `caffeine_builtin_alloc_symbolic` to allocate a fixed array of symbolic bytes when the size is a reasonably-sized concrete value.

This PR is part of #335 

cc @brookedolny The changes to `Context` in this PR will probably affect the code you'll have to write for writing test cases to the serialized representation.

### DIFF
--- a/src/Interpreter/Interpreter.cpp
+++ b/src/Interpreter/Interpreter.cpp
@@ -626,7 +626,7 @@ std::optional<std::string> readSymbolicName(std::shared_ptr<Solver> solver,
   const char* start = values + offset;
   const char* end = (const char*)std::memchr(start, 0, size - offset);
   if (!end) {
-    CAFFEINE_UNSUPPORTED("Symbolic name was not nul-terminated");
+    CAFFEINE_UNSUPPORTED("Symbolic name was not null-terminated");
     return std::nullopt;
   }
 
--- a/src/Interpreter/Interpreter.cpp
+++ b/src/Interpreter/Interpreter.cpp
@@ -19,6 +19,12 @@
 
 namespace caffeine {
 
+namespace {
+  // The maximum size for which a fixed-size symbolic constant will be optimized
+  // to a fixed array of smaller constants.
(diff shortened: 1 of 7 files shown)


src/Interpreter/Interpreter.cpp
```diff
@@ -616,15 +623,14 @@ std::optional<std::string> readSymbolicName(std::shared_ptr<Solver> solver,
...
 
   const char* values = array.data();
-  for (size_t i = offset; i < size; ++i) {
-    if (values[i] == 0)
-      return name;
-    name.push_back(values[i]);
+  const char* start = values + offset;
+  const char* end = (const char*)std::memchr(start, 0, size - offset);
+  if (!end) {
+    CAFFEINE_UNSUPPORTED("Symbolic name was not nul-terminated");
```
Skrinikov, line 629: was it supposed to be `null-terminated`?

src/Interpreter/PrintingFailureLogger.cpp
```diff
@@ -103,12 +63,18 @@ void PrintingFailureLogger::log_failure(const Model& model, const Context& ctx,
   ss << "Found assertion failure:\n";
 
   if (model.result() == SolverResult::SAT) {
-    ConstantPrinter printer{ss, &model};
+    // ConstantPrinter printer{ss, &model};
```
Skrinikov, line 66: Why is this commented out?

PR #384, opened 2021-05-08
Add the ability to checkpoint and restore an `AssertionList`.

Phantomical:
This PR adds two helper functions to `AssertionList`.
- `checkpoint` returns a value that can be then fed into
- `restore` which removes all values that have been inserted into the list since the corresponding `checkpoint` call.

This is useful as a general "insert things into the list to see if they're already present" strategy. In this PR, I've used it to simplify `Z3Solver`.

At the same time I ran into bugs in `AssertionList` and `SparseVector` since checkpointing/restoring results in them being in new states. I've included fixes and a test case in this PR as well.


Skrinikov (approved):
lgtm

PR #387, opened 2021-05-10
Implement arbitrary tracing support

Phantomical:
This PR adds a small tracing library that can be used to log arbitrary spans while caffeine is running. This is meant to allow for easier performance analysis. I've already used it to diagnose one performance improvement (#386).

If disabled at compile time, the tracing support should optimize away to nothing. If not, it has to check one global variable to see if it's enabled. I've put the existing tools I've written to work with it [here](https://github.com/insufficiently-caffeinated/tracing-tools).

## Details
This required a few changes to more than just setting up tracing. Here are the detailed changes:
- We now have a config header that is configured based on cmake options passed to the cmake configure step. The template for this is `Config.h.in` in the root of the repo.
- There is now a `CAFFEINE_ENABLE_TRACING` option to disable tracing at compile time
- I've added a capnp message for serialized trace data
- `caffeine/Support/Tracing.(h|cpp)` contains all the tracing code

I've also set up a global trace handler in the main binary as well as some recording of trace spans for calls to the Z3 solver.

### DIFF
--- a/src/Support/Tracing.cpp
+++ b/src/Support/Tracing.cpp
@@ -1,0 +1,162 @@
+#include "caffeine/Support/Tracing.h"
+#include "caffeine/Protos/tracepoint.capnp.h"
+#include "caffeine/Support/Assert.h"
+#include <capnp/message.h>
+#include <capnp/orphan.h>
+#include <capnp/serialize-packed.h>
+#include <chrono>
+#include <cstdio>
+#include <fmt/format.h>
+#include <fmt/ostream.h>
+#include <llvm/ADT/SmallVector.h>
+#include <thread>
+#include <vector>
+
+namespace caffeine::tracing {
+
+namespace detail {
+  void annotate_tid(AutoTraceBlock& block) {
+    block.annotate("tid", fmt::format("{}", std::this_thread::get_id()));
+  }
+} // namespace detail
+
+#if CAFFEINE_ENABLE_TRACING
+
+namespace {
+  static TraceSink* trace_sink = nullptr;
+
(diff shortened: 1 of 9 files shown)


src/Support/Tracing.cpp
```diff
@@ -0,0 +1,162 @@
...
+      list[i].adoptValue(std::move(value));
+
+      i += 1;
+    }
+
+    kj::VectorOutputStream stream;
+    capnp::writePackedMessage(stream, message);
+
+    auto data = stream.getArray();
+    fwrite(data.begin(), data.size(), 1, sink->file());
```
mishazharov, line 102: Capnp can also directly write to a file descriptor with something like `writePackedMessageToFd` which would likely be more efficient than buffering it first?

Since we're dealing with `FILE*`s here we might need to check if we're on a posix and convert it to an FD, or we can implement something like this https://github.com/capnproto/capnproto/issues/46 (essentially https://stackoverflow.com/a/47048097 but with a subclass of `kj::OutputStream` that we create)

mishazharov (approved):
Minor nit, otherwise LGTM

src/Support/Tracing.cpp
Phantomical: The problem with this approach is that it doesn't work when multiple threads are writing to the fd. You'd get multiple messages being interleaved.

`fwrite` is guaranteed to be atomic on all the platforms we care about so thats why I use it here.

mishazharov: True, in that case LGTM!

PR #398, opened 2021-05-15
AFL mutator integration

mishazharov:
This draft PR is my attempt at getting caffeine to generate testcases automatically for AFL. Some fun things to note:
* I'm not sure if it's legal to `write_to_testcase` in a mutator
* It's currently untested ¯\\_(ツ)_/¯


CMakeLists.txt
```diff
@@ -77,7 +77,7 @@ if (CAFFEINE_ENABLE_BUILD)
 
   # LLVM Requires an exact version specification so we need to do one
   # find_package call for each LLVM version we want to support
-  find_package(LLVM 11.1)
+  find_package(LLVM 11.1 QUIET)
```
Phantomical, line 80: Could you put this change into a separate PR?

Dockerfile
```diff
@@ -29,4 +29,3 @@ RUN apt-get update \
         pkg-config \
     && update-alternatives --install /usr/local/bin/llvm-config llvm-config /usr/bin/llvm-config-11 20 \
     && rm -rf /var/lib/apt/lists/*
-RUN update-alternatives --install /usr/local/bin/llvm-config llvm-config /usr/bin/llvm-config-11 20
```
Phantomical, line 32: Same thing here.

include/caffeine/Support/DiagnosticHandler.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+#include <llvm/IR/Module.h>
+
+namespace caffeine {
+
+struct DecafDiagnosticHandler : public llvm::DiagnosticHandler {
```
Phantomical, line 7: I'd also pull this refactoring into a separate PR and then (optionally) stack this PR on top of that one.

src/Interpreter/Interpreter.cpp
```diff
@@ -42,7 +42,8 @@ void Interpreter::logFailure(Context& ctx, const Assertion& assertion,
   if (result != SolverResult::SAT)
     return;
 
-  logger->log_failure(result.model(), ctx, Failure(assertion, message));
+  if (logger)
+    logger->log_failure(result.model(), ctx, Failure(assertion, message));
```
Phantomical, line 46: Question: Under what case would we end up with a null logger? Maybe we should just have an assertion when building the interpreter?

src/bin/caffeine.cpp
```diff
@@ -66,36 +66,7 @@ cl::opt<std::string> enable_tracing{
...
-    DiagnosticPrinterRawOStream dp(errs());
-    di.print(dp);
-    errs() << '\n';
-
-    return true;
-  }
-};
-
-} // namespace
+namespace {} // namespace
```
Phantomical, line 69:
```suggestion
```

tools/guided-fuzzing/src/CaffeineMutator.cpp
```diff
@@ -0,0 +1,165 @@
...
+#include "caffeine/Interpreter/Policy.h"
+#include "caffeine/Interpreter/Store.h"
+#include "caffeine/Solver/CanonicalizingSolver.h"
+#include "caffeine/Solver/SequenceSolver.h"
+#include "caffeine/Solver/SimplifyingSolver.h"
+#include "caffeine/Solver/Z3Solver.h"
+#include "caffeine/Support/DiagnosticHandler.h"
+
+#include "include/GuidedExecutionPolicy.h"
+#include "include/TestCaseFailureLogger.h"
```
Phantomical, line 21: Maybe just add `tools/guided-fuzzing/include` to the include path? Having "include" as part of header includes is a bit weird.

tools/guided-fuzzing/src/GuidedExecutionPolicy.cpp
```diff
@@ -17,7 +22,22 @@ bool GuidedExecutionPolicy::should_queue_path(const Context& ctx) {
...
+  // If the path that we're checking isn't SAT given the assumptions, we want
+  // to see if it's SAT in general without our additional assumption so that
+  // we can create a testcase out of it. The key here is that the
+  // `NeverQueuePolicy` will turn our `ctx_copy` into a testcase after it is
+  // done running (assuming it is SAT)
+  auto store = NopStore();
+  auto policy = NeverQueuePolicy(cases, mutator);
+  auto ctx_copy = ctx;
+  Interpreter interp(&ctx_copy, &policy, &store, nullptr, mutator->solver);
+  interp.execute();
```
Phantomical, line 38: I don't think this will work properly. Maybe you wanted this instead? It seems to do what you want.
```suggestion
  // If the path that we're checking isn't SAT given the assumptions, we want
  // to see if it's SAT in general without our additional assumption so that
  // we can create a testcase out of it. 
  auto policy = NeverQueuePolicy(cases, mutator);
  policy.on_path_complete(ctx, ExitStatus::Dead);
```

Phantomical (commented):
Not sure if you wanted a review yet so I haven't reviewed too deeply. Looks pretty good to me overall though :)

I think you should pull a few of the changes out into separate PRs. All the ones I've left comments on should be easy stamp approvals and that would make it easier to review this PR.

Other than that I've left a few comments on things I found that looked weird to me.

tools/guided-fuzzing/src/GuidedExecutionPolicy.cpp
mishazharov: The reason I'm creating a new interpreter is because I want to see if the `Context` being passed in is `SAT` when executed (without additional assertions).

mishazharov: After some offline discussion I see your point. Also created #401 to fix some issues related to this

Phantomical: Creating a new interpreter would continue executing along the path. I'm going to try and draw this out with some diagrams:

So suppose the main path through the program looks sorta like this (horizontal for compactness)
```
____________________________________________________________
```
and then we have branches that go off but aren't the same path as taken by the test case
```
____________________________________________________________
    \                           \                       \
```

If we don't create a new interpreter we're just creating test cases for those branches that go off.

However, if we create a new interpreter and keep executing we're doing something like this:
```
____________________________________________________________
      \                   \                         \____________________
       \                   \_________                    \            \
        \                       \
         \__________
             \     \
```
That is, we don't just follow the main test case, we also follow every branch off of it until the next branch.

What I think you want is the first one. Continuing executing branches that aren't the main one doesn't tell if you anything more about whether it's currently SAT than just running the solver over the current path condition (`ctx.assertions`). The only exception to this is if you have a failure via `logger->log_failure` in which case you also need to include the extra assertion that gets passed to log_failure.

Phantomical (approved):
Left some comments, LGTM otherwise but you'll probably want to address them in a follow-up commit.

src/Interpreter/Store.cpp
```diff
@@ -63,6 +63,10 @@ void QueueingContextStore::shutdown() {
   condvar.notify_all();
 }
 
+QueueingContextStore::~QueueingContextStore() {
+  shutdown();
+}
```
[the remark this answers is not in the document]
mishazharov: Nah, was debugging something and trying to narrow down suspects. But you're right even if this was the issue the fix wouldn't have helped

PR #399, opened 2021-05-16
Fix error when changing generators without removing all files

Phantomical:
If you change the generator for the cmake build directory without deleting all the files in the build directory then you'd run into errors for the packages downloaded by caffeine_dependency as they would still be using the old generator. This fixes that by deleting them if the generator has changed since the last time cmake ran.

Ultimately, it's a pretty rare error so it doesn't matter too much if it gets fixed but it's the right thing to do.


mishazharov (approved):
Nice!

PR #403, opened 2021-05-16
Extract DiagnosticHandler from caffeine bin

mishazharov:
As part of decomposing #398

### DIFF
--- a/include/caffeine/Support/DiagnosticHandler.h
+++ b/include/caffeine/Support/DiagnosticHandler.h
@@ -1,0 +1,11 @@
+#pragma once
+
+#include <llvm/IR/Module.h>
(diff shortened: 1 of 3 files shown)


include/caffeine/Support/DiagnosticHandler.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+#include <llvm/IR/Module.h>
```
Phantomical, line 3: Is this the right include for this? Shouldn't there be one for `llvm::DiagnosticHandler` itself?

Phantomical (approved):
One comment regarding includes but LGTM otherwise

PR #405, opened 2021-05-17
Don't use cmake -E rm for older cmake versions

Phantomical:
As in title. Previously this caused a non-fatal error to be printed out on cmake versions older than 3.17.


mishazharov (approved):
Nice

PR #407, opened 2021-05-19
Implement `std::unordered_map` variant with `std::weak_ptr` keys

Phantomical:
This PR adds a new container `weak_map` that acts as an `std::unordered_map` with `std::weak_ptr` keys. That is, it's somewhat like a cache that automatically cleans up values when the corresponding `std::shared_ptr` key is deleted (eventually). I will be using this to implement caching within a new solver that will be introduced in the next set of PRs.

The code added here is pretty straightforward -- if somewhat verbose -- because it can forward to the underlying `std::unordered_map` for everything.

### DIFF
--- a/include/caffeine/ADT/WeakMap.h
+++ b/include/caffeine/ADT/WeakMap.h
@@ -1,0 +1,252 @@
+#pragma once
+
+#include "caffeine/ADT/SparseVector.h"
+#include <boost/range/adaptor/filtered.hpp>
+#include <boost/range/adaptor/transformed.hpp>
+#include <memory>
+#include <unordered_map>
+#include <utility>
+
+namespace caffeine {
+
+/**
+ * A hash map that uses std::weak_ptr as keys. Once the weak_ptr expires it will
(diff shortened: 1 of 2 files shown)


include/caffeine/ADT/WeakMap.h
```diff
@@ -0,0 +1,252 @@
...
+
+    if (map.size() == prune_size)
+      prune();
+
+    return {iterator(it2), inserted};
+  }
+
+  std::pair<iterator, bool> insert(const std::shared_ptr<T>& key,
+                                   const V& value) {
+    return emplace(key, value);
```
mishazharov, line 175: Should it `prune` on insertions as well?

mishazharov (approved):
LGTM! Non blocking nit in the comments

include/caffeine/ADT/WeakMap.h
Phantomical: It does, it just happens within `emplace`.

PR #410, opened 2021-05-19
Implement string interning for `Symbol` names

Phantomical:
This PR makes the `Symbol` constructor transparently intern the string name when applicable. The motivation for this is that there are other PRs which are going to involve manipulating a lot of symbols and this should both speed them up and reduce memory usage.

The main benefits of this are:
- Checking symbols for equality is now O(1)
- Hashing symbols is now O(1)
- Symbols are now slightly smaller in size

The main disadvantage is that we can never free strings unless we know that there are no symbols still in existance. In the future I may go and add some sort of reference counting to avoid this problem.


include/caffeine/ADT/StringInterner.h
```diff
@@ -0,0 +1,88 @@
...
+} // namespace caffeine
+
+namespace std {
+template <>
+struct hash<caffeine::InternedString> {
+  size_t operator()(const caffeine::InternedString& s) const noexcept {
+    return hash<const char*>()(s.data());
+  }
+};
+} // namespace std
```
mishazharov, line 88: \n?

mishazharov (approved):
Small nit, otherwise LGTM

PR #411, opened 2021-05-19
Add solver which only includes assertions needed to solve the unproven ones

Phantomical:
This is based off of the Green paper (see comment on `ConstraintSlicer` for the full title) and basically does exactly what the title says.


src/Query/ConstraintSlicer.cpp
```diff
@@ -0,0 +1,89 @@
+#include "caffeine/Query/ConstraintSlicer.h"
+#include "caffeine/Interpreter/AssertionList.h"
+
+namespace caffeine {
+
+AssertionList ConstraintSlicer::slice(const AssertionList& assertions,
+                                      const Assertion& extra) {
+  /**
+   * The algorithm used here is quite simple, we just add all the that share
```
mishazharov, line 9: `all the that`?

assertions?

mishazharov (approved):
found a typo but otherwise LGTM

PR #417, opened 2021-05-27
Implement `extractvalue` and `insertvalue` instructions

Phantomical:
These instructions allow you to interact with literal struct fields within LLVM. They're not used by most code but code generated as a result of some builtins makes use of them.

### Full Changes
- Implement `extractvalue`/`insertvalue` in `ExprEvaluator`
- Add non-const accessors to `LLVMValue` so that they can be used for `insertvalue`
- Add two test cases covering the easy case for the implementation

### DIFF
--- a/test/run-pass/inst/extractvalue/nested.ll
+++ b/test/run-pass/inst/extractvalue/nested.ll
@@ -1,0 +1,19 @@
+target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128"
+target triple = "x86_64-pc-linux-gnu"
+
+define void @test(i32 %x) {
+  %a = insertvalue  {{{{{{{ i32 }}}}}}} undef, i32 %x, 0, 0, 0, 0, 0, 0, 0
+  %c = extractvalue {{{{{{{ i32 }}}}}}} %a,            0, 0, 0, 0, 0, 0, 0
+  %d = icmp eq i32 %c, %x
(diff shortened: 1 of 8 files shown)


test/run-pass/inst/extractvalue/nested.ll
```diff
@@ -0,0 +1,19 @@
+target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128"
+target triple = "x86_64-pc-linux-gnu"
+
+define void @test(i32 %x) {
+  %a = insertvalue  {{{{{{{ i32 }}}}}}} undef, i32 %x, 0, 0, 0, 0, 0, 0, 0
```
Skrinikov, line 5: ok, first time seeing this in llvm why `{{{{{{{`?

Skrinikov (approved):
everything else looks good! 
Im just curious about previous comment

test/run-pass/inst/extractvalue/nested.ll
Phantomical: This is a test to see whether inserting/extracting to/from nested structs work as expected. In LLVM IR `{ i32 }` is a struct containing an `i32`, `{ { i32 } }` is a struct containing a `{ i32 }` and so on. So what we have here is 7 layers of structs around an `i32`.

Skrinikov: gotcha. Looks good to me then

PR #419, opened 2021-05-28
Add coverage reports to CI

mishazharov:
This PR attempts to add coverage reports to CI using codecov

### DIFF
--- a/.github/scripts/setup-env.sh
+++ b/.github/scripts/setup-env.sh
@@ -22,11 +22,11 @@ done
 
 case $COMPILER in
   clang)
-    export CC=clang
-    export CXX=clang++
+    echo "CC=clang" >> $GITHUB_ENV
+    echo "CXX=clang++" >> $GITHUB_ENV
     ;;
   gcc)
-    export CC=gcc
-    export CXX=g++
+    echo "CC=gcc" >> $GITHUB_ENV
+    echo "CXX=g++" >> $GITHUB_ENV
     ;;
 esac
--- a/.github/scripts/setup-env.sh
+++ b/.github/scripts/setup-env.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # This script is meant to simplify the compilation matrix in the case where we have
 # multiple dependent definitions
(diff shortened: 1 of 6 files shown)


.github/scripts/setup-env.sh
```diff
@@ -21,11 +22,11 @@ done
...
-    export CC=clang
-    export CXX=clang++
+    echo "CC=clang-11" >> $GITHUB_ENV
+    echo "CXX=clang++-11" >> $GITHUB_ENV
     ;;
   gcc)
-    export CC=gcc
-    export CXX=g++
+    echo "CC=gcc" >> $GITHUB_ENV
+    echo "CXX=g++" >> $GITHUB_ENV
```
mishazharov, line 30: It turns out we never the compiler properly and have always been using GCC (for instance see https://github.com/insufficiently-caffeinated/caffeine/runs/2687228648 )

.github/workflows/find-coverage.yml
```diff
@@ -0,0 +1,55 @@
...
+
+      - name: Coverage
+        env:
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+        run: |
+          cd build
+          find . -name "*.profraw"
+          llvm-profdata-11 merge -sparse `find . -name "*.profraw"` -o default.profdata
+          llvm-cov-11 show -instr-profile ./default.profdata bin/caffeine-unittest -format=text > coverage.txt
+          apt update && apt install curl # this is a bandaid until we get curl in the container
```
mishazharov, line 53: Temporary hack

.github/scripts/setup-env.sh
```diff
@@ -7,10 +7,11 @@ error() {
   echo "$1" >&2
 }
 
-while [[ $# -gt 0 ]]; do
+while [ $# -gt 0 ]; do
   case $1 in
     --compiler)
       COMPILER="$2"
+      break
```
Phantomical, line 14: This will make this work better once we have more options
```suggestion
      shift
      shift
```

.github/workflows/find-coverage.yml
Phantomical: I wouldn't worry about this.

Phantomical (approved):
One suggestion otherwise LGTM

PR #422, opened 2021-05-29
Change command line entry point to no longer be a mandatory argument

Phantomical:
Previously the entry point was a mandatory positional argument. However, we almost always want to use `main` as the entry point so it doesn't make sense for it to be a required argument. This PR changes it to be a flag `--entry` with the default value of `main`.

I've also removed the code that will automatically generate symbolic constants for the arguments of the entry point method. I have, however, left the special casing for `main` so that existing tests continue to work.

### DIFF
--- a/src/Interpreter/Context.cpp
+++ b/src/Interpreter/Context.cpp
@@ -30,39 +30,18 @@ Context::Context(llvm::Function* function)
   stack.emplace_back(function);
   StackFrame& frame = stack_top();
 
-  if (function->getName() == "main") {
-    const llvm::DataLayout& layout = mod->getDataLayout();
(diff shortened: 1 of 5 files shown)


src/Interpreter/Context.cpp
```diff
@@ -12,57 +12,50 @@
...
-  }
-
-  std::string message;
-  llvm::raw_string_ostream os{message};
-  os << "Unsupported LLVM type: ";
-  type->print(os);
-
-  CAFFEINE_ABORT(message);
+Context::Context(llvm::Function* function,
+                 const std::unordered_map<llvm::Value*, OpRef>& args)
```
mishazharov, line 16: Nice addition! Should document why this is useful though

mishazharov (approved):

PR #425, opened 2021-05-30
Fix some issues with test coverage generation

Phantomical:
Previously the generated `default.profraw` file for all the test cases would all overwrite each other when being executed. In larger parallel builds this usually led to corrupted coverage info. Even if it worked it would result in the reports missing data for the other test cases.

This fixes that by doing a couple of different things:
1. When coverage is enabled we run caffeine-unittest all at once instead of generating a different cmake test case for each contained test.
2. We use the `LLVM_PROFDATA_FILE` env var to make the run-file tests generate their profiling data with a different filename for  test.

There's still some holes in this (googletest death tests don't properly generate coverage at the moment) but it should be better than the status quo.


mishazharov (approved):
Noice

PR #428, opened 2021-06-03
Various fixups to afl-guided-mutator

Phantomical:
I spent at least 6 hours debugging why this didn't work only to discover it was mainly due to different optimization settings when building the test case. However I also made a bunch of changes that I felt were worth having.

## Detailed List of Changes
- Add conversions between `Span` and `std::string_view`. (not really a part of the change but I don't really want to make a separate PR for this)
- Test cases are now stored as `std::string` instances instead of `SharedArray`
- There is now a separate entry point `__caffeine_entry_point` instead of calling `main`.  This allows us to set the size of the symbolic constant appropriately without having to resort to a symbolic size. It also means that it should be able to be linked into a real program without too much of an issue.
- Put the slicing solver into the solver stack. (We should probably have a builder for this)
- Moved test case generation from `should_queue_path` to `on_path_complete` since it's guaranteed to be called for every path.
- Added a check that prevents us from performing an empty mutation since just returning the same test case really isn't that useful.
- Fixed a memory leak due to allocating a buffer with `malloc` and assuming that AFL frees it when it does not.

### DIFF
--- a/tools/guided-fuzzing/src/CaffeineMutator.cpp
+++ b/tools/guided-fuzzing/src/CaffeineMutator.cpp
@@ -8,6 +8,7 @@
 #include <llvm/IRReader/IRReader.h>
 #include <llvm/Support/CommandLine.h>
 
+#include "caffeine/ADT/Guard.h"
 #include "caffeine/ADT/Span.h"
 #include "caffeine/Interpreter/Interpreter.h"
 #include "caffeine/Interpreter/Policy.h"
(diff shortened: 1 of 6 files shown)


tools/guided-fuzzing/src/CaffeineMutator.cpp
```diff
@@ -75,32 +84,28 @@ size_t CaffeineMutator::mutate(caffeine::Span<char> data) {
...
-  auto& test_case = cases->at(cases->size() - 1);
+    if (last_case.size() > max_size)
+      break;
 
-  *out_buf = (unsigned char*)malloc(test_case.size());
-  memcpy(*out_buf, test_case.data(), test_case.size());
-
-  auto len = test_case.size();
-  cases->pop_back();
+    *out_buf = (unsigned char*)last_case.data();
```
mishazharov, line 94: Nice catch with the `malloc`

mishazharov (approved):
Thanks for helping debug this!

PR #447, opened 2021-06-09
Add install targets for caffeine

Phantomical:
As in title. This allows us to run `make install` to install caffeine and then other projects can use `find_package(caffeine)` to make use of caffeine. The current targets that are included are
- `caffeine`
- `caffeine-opt-plugin`
- `caffeine-bin`
- All public headers (ones in `include` and generated ones)
- The builtins and libc bitcode libraries
- vcpkg triplet files

I haven't included the afl mutator since it doesn't match the naming conventions (it should be named caffeine-... so that it doesn't conflict with other libraries under `lib`)

I'm still not quite sure how to expose bitcode libraries. For the moment I've elected to just put them under `lib/caffeine` and add some variables that have the file names.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -76,50 +76,6 @@ mark_as_advanced(CAFFEINE_FMTONLY)
 # Disable messages for targets on makefile generators
 set_property(GLOBAL PROPERTY TARGET_MESSAGES OFF)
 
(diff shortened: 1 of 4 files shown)


mishazharov (approved):
Nice!!

brookedolny (approved):

PR #449, opened 2021-06-15
Autogen `__caffeine_entry_point` for AFL++ mutator

mishazharov:
This PR adds a default implementation of `__caffeine_entry_point` in case the user program does not provide it. Essentially, it adds
```
void __caffeine_entry_point(int size) {
  uint8_t data[size] = {0};
  caffeine_make_symbolic(data, size, "__caffeine_mut");
  LLVMFuzzerTestOneInput(data, size);
}
```
to the module.

### DIFF
--- a/tools/guided-fuzzing/src/CaffeineMutator.cpp
+++ b/tools/guided-fuzzing/src/CaffeineMutator.cpp
@@ -25,8 +25,8 @@
 #include "GuidedExecutionPolicy.h"
 
 #define CAFFEINE_FUZZ_TARGET "LLVMFuzzerTestOneInput"
-#define CAFFEINE_FUZZ_START "__caffeine_entry_point"
-#define CAFFEINE_MAKE_SYMBOLIC "caffeine_make_symbolic"
+#define CAFFEINE_FUZZ_START "caffeine.entry_point"
+#define CAFFEINE_MAKE_SYMBOLIC "caffeine_builtin_symbolic_alloca"
 
 namespace caffeine {
 class NullFailureLogger : public caffeine::FailureLogger {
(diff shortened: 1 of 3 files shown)


tools/guided-fuzzing/src/CaffeineMutator.cpp
```diff
@@ -44,10 +52,56 @@ CaffeineMutator::CaffeineMutator(std::string binary_path, afl_state_t* afl) {
...
 
+  auto bitwidth = this->module->getDataLayout().getPointerSizeInBits();
+
+  // Create CAFFEINE_FUZZ_START automatically
   fuzz_target = module->getFunction(CAFFEINE_FUZZ_START);
-  if (fuzz_target == nullptr) {
-    llvm::WithColor::error() << "No method '" << CAFFEINE_FUZZ_START << "'\n";
-    CAFFEINE_ABORT();
+  if (!fuzz_target) {
+    fuzz_target = llvm::Function::Create(
```
mishazharov, line 60: TODO: pull this out into a function before merging

tools/guided-fuzzing/src/CaffeineMutator.cpp
```diff
@@ -25,8 +26,71 @@
...
+    if (caffeine_make_symbolic == nullptr) {
+      llvm::WithColor::error()
+          << "No method '" << CAFFEINE_MAKE_SYMBOLIC << "'\n";
+      CAFFEINE_ABORT();
+    }
+
+    auto bb = llvm::IRBuilder{
+        llvm::BasicBlock::Create(*llvm_context, "body", fuzz_target)};
+
+    auto alloca = bb.CreateAlloca(llvm::Type::getInt8PtrTy(*llvm_context),
```
Phantomical, line 67: I think this should be `getInt8Ty` right? Otherwise we're creating an array of pointers. This will work (since `sizeof(i8) < sizeof(i8*)` but I don't think it's what we want.

tools/guided-fuzzing/src/CaffeineMutator.cpp
```diff
@@ -25,8 +26,71 @@
 
 #define CAFFEINE_FUZZ_TARGET "LLVMFuzzerTestOneInput"
 #define CAFFEINE_FUZZ_START "__caffeine_entry_point"
```
Phantomical, line 28: Since this is builtin maybe we should change it to something that won't collide with anything from C such as `caffeine.entry_point`?

Phantomical (approved):
Approving with some reservations - the code you've got here will work although it'll perform an unnecessary copy and will fail if the test program doesn't link to the runtime.

For a bit more explanation, you're currently generating something like this (pseudocode since I don't recall the exact semantics)

```llvm
define void __caffeine_entry_point(i64 %size) {
  %a = alloca [i8 x %size]
  call void @caffeine_make_symbolic(i8* %a, i64 %size, "__caffeine_mut")
  call void @LLVMFuzzerTestOneInput(i8* %a, i64 %size)
}
```

However, it would be more efficient (and work without the builtins library) if you generated
```llvm
define void __caffeine_entry_point(i64 %size) {
  %a = call i8* @caffeine_builtin_symbolic_alloca(i64 %size, "__caffeine_mut")
  call void @LLVMFuzzerTestOneInput(i8* %a, i64 %size)
}
```
as it doesn't have to do a copy to an allocation of the exact same size. Furthermore, `caffeine_make_symbolic` is defined in the builtins library and the program might not have it linked. `caffeine_builtin_symbolic_alloca` literally just directly returns the symbolic allocation so it's way faster than `caffeine_make_symbolic` and beyond that it's a primitive exported by caffeine so it'll always be available.

tools/guided-fuzzing/src/CaffeineMutator.cpp
mishazharov: Nah, I think `getInt8PtrTy` creates a pointer to an `Int8` (which can be an array I guess). So I think it makes sense since the prototype needs an int pointer?

PR #452, opened 2021-06-20
Add readme for Caffeine mutator

mishazharov:
This is a simple PR which adds a bit of documentation about using the caffeine mutator for AFL++


Phantomical (approved):
LGTM!

PR #456, opened 2021-06-21
Add utilities for implementing forking operations in builtins

Phantomical:
A lot of operations involve first performing an operation that could fork and then doing other things on all the resulting contexts. This ends up being very tedious to write (e.g. resolving a pointer requires ~50 LOC every time we want to do it). This PR adds a utility that handles tracking all the forks and executing a series of operations on them.

So far I've only added the operations that were used in implementing `setjmp` and `longjmp`. As more operations are needed then they can be added. In the meantime, they will all be wrappers around the `transform_fork` function so there's no functionality that couldn't be implemented manually.


mishazharov (approved):
LGTM! Nice rewrite

PR #463, opened 2021-07-05
Add `go` and `gllvm` to Dockerfile

mishazharov:
These two dependencies are going to be useful to c++ support in our test
suite.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,5 +27,9 @@ RUN apt-get update \
         capnproto \
         pkg-config \
         curl \
-    && update-alternatives --install /usr/local/bin/llvm-config llvm-config /usr/bin/llvm-config-11 20 \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -27,5 +27,9 @@ RUN apt-get update \
         capnproto \
         pkg-config \
         curl \
-    && update-alternatives --install /usr/local/bin/llvm-config llvm-config /usr/bin/llvm-config-11 20 \
     && rm -rf /var/lib/apt/lists/*
+RUN update-alternatives --install /usr/local/bin/llvm-config llvm-config /usr/bin/llvm-config-11 20 \
```
mishazharov, line 31: I also unwrapped this install command because it makes rebuilds faster which is extremely useful when testing locally

Phantomical (approved):

PR #464, opened 2021-07-10
Implement better multithreaded context store

### LINKED ISSUE #459: Implement more efficient `ExecutionContextStore`
The current `ExecutionContextStore` instance ends up serializing everything through a single queue. This has two main problems
- Everything hitting the queue ends up being serialized behind a single lock
- Lots of spinning as new threads come in and end up blocked on the lock
- Little to no cache reuse since usually work will be sent to another thread in the threadpool

The new `ExecutionContextStore` implementation should address these issues.

## Existing work for Inspiration
- One approach would be to do it in the same way as [TBB](https://github.com/oneapi-src/oneTBB) manages tasks for it's threadpool, AFAIK this is done by having a deque per thread. The current thread will pull/push from the front of it's deque while other threads pull from the back when they need more work.

Phantomical:
Currently we use `QueueingContextStore` (or a variant thereof that caches the last value). However, this has a problem that if there isn't enough work to saturate all the worker threads then the unused threads will spend a bunch of their time bouncing through locks when new values are inserted into the queue. This results in caffeine using far more CPU than is useful. Furthermore, when new values are inserted generally a different thread will win the race to pull the context out of the store. This means that there is little cache locality when dealing with these contexts.

This PR introduces a new context store called `ThreadQueueContextStore` that keeps a separate deque for each thread. When a thread needs to get a new context it'll pull the most recent one off of it's own queue. If it has no work items then it will pull the oldest element of the queue of a random thread. If unable to find an element in the queue then it will sleep for a period before trying again. This ensures that inactive threads use little CPU while also ensuring that they find work in a reasonable amount of time.

This PR also includes a change to `ThreadMap` so that it can be used with non-default-initializable types as well as some small cleanups within the `Tracing.h`. Furthermore, I've deleted the old `ThreadQueuedExecutor` since it is worse in every single use case compared to the new one.

Closes #459

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -108,7 +108,7 @@ endif()
 # TODO: We should have an option to download and build these locally if needed.
 find_package(GTest REQUIRED)
 find_package(Z3 4.8.7 REQUIRED)
-find_package(Boost REQUIRED)
+find_package(Boost COMPONENTS thread REQUIRED)
 find_package(fmt REQUIRED)
 find_package(CapnProto REQUIRED)
 
--- a/include/caffeine/ADT/ThreadMap.h
+++ b/include/caffeine/ADT/ThreadMap.h
@@ -16,7 +16,7 @@ public:
     auto lock = std::shared_lock(mutex);
 
     auto it = map.find(id);
-    return it == map.end() ? nullptr : &it->second;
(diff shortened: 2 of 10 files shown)


mishazharov (approved):
LGTM, nice!

PR #466, opened 2021-07-12
Rework `Operation` cache to avoid blocking threads

Phantomical:
This PR does a few things
1. Add a new dependency on https://github.com/Tessil/hopscotch-map which is a fast hashmap library.
2. Rework the existing operation cache to use a shared lock so that no threads get blocked if the operation is already in the cache.
3. Move the operation cache to its own public header.
4. Change the last few non-cached operation types to be fully cached now.

The one exception here is that `FixedArray` is still lazily cached (i.e. only cached when we need to build an operation that uses it directly). This is to avoid quadratic blowup when caching lots of temporary writes to an array.

This is part of a series of PRs cleaning up caffeine in preparation for focusing more on optimization.

### DIFF
--- a/src/IR/OperationCache.cpp
+++ b/src/IR/OperationCache.cpp
@@ -1,0 +1,5 @@
+#include "caffeine/IR/OperationCache.h"
+
+namespace caffeine {
+  
+}
--- a/src/IR/OperationCache.cpp
+++ b/src/IR/OperationCache.cpp
@@ -1,5 +1,116 @@
 #include "caffeine/IR/OperationCache.h"
+#include "caffeine/Support/Assert.h"
+
+#include <algorithm>
 
 namespace caffeine {
-  
+
+OperationCache* OperationCache::default_cache() {
+  static OperationCache cache;
+  return &cache;
 }
+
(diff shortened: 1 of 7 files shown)


src/IR/OperationCache.cpp
```diff
@@ -0,0 +1,116 @@
...
+void OperationCache::gc_locked(std::unique_lock<std::shared_mutex>&) {
+  for (auto it = set.begin(), end = set.end(); it != end;) {
+    if (it->use_count() == 1) {
+      it = set.erase(it);
+    } else {
+      ++it;
+    }
+  }
+
+  threshold = std::max(set.size() * 2, min_threshold);
```
mishazharov, line 113: The threshold just keeps growing infinitely? Maybe we want a max as well?

mishazharov (approved):
One nit, otherwise LGTM!

src/IR/OperationCache.cpp
Phantomical: The threshold will shrink if lots of the values in the set just got pruned. If there's lots of dead values then we should be able to shrink quite a bit.

PR #468, opened 2021-07-13
Add CI workflow that doesn't use the docker container

Phantomical:
This PR adds an additional actions workflow that doesn't use the docker image. Instead it
- Takes advantage of the pre-installed libraries already on the ubuntu-latest github actions image
- Should (maybe) cache the build directory for faster follow-on rebuilds
- Uses ninja to build in CI (no more nonsensical percentage markers)
- Should be at least 30s faster due to not having to download an entire OS
- Also caches the output build directory for (hopefully) faster rebuilds.

The combination of caching + apt dependencies looks like it is able to bring the build time down to 1-2 minutes for a cached build.

### DIFF
--- a/.github/workflows/build-apt.yml
+++ b/.github/workflows/build-apt.yml
@@ -1,0 +1,73 @@
+name: build-apt
+
+on:
+  # Don't build CI for stackbot internal branches, they are already built for PRs
+  push:
+    branches-ignore:
+      - stackbot/**
+  pull_request:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        compiler: [clang, gcc]
+      # We explicitly want to run all the compilers
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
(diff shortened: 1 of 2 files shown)


.github/workflows/build-apt.yml
```diff
@@ -0,0 +1,75 @@
...
+        run: |
+          sudo apt-get install -y \
+            ninja-build \
+            libgtest-dev \
+            libfmt-dev \
+            libboost-all-dev \
+            capnproto libcapnp-dev \
+            git-restore-mtime
+          sudo ./.github/scripts/update-alternatives-llvm.sh 11 20
+          go get github.com/SRI-CSL/gllvm/cmd/...
```
mishazharov, line 35: Minor nit: does this ubuntu install actually have the go bin directory in the path (where go places user binaries)? If you don't know that's fine

mishazharov (approved):
LGTM, have a question but feel free to merge

.github/workflows/build-apt.yml
Phantomical: No clue - I'm assuming yes but we don't use this anywhere.

PR #469, opened 2021-07-13
Convert coverage CI setup to avoid using containers

Phantomical:
This applies the same changes as in #468 to the coverage workflow. The one exception is that it doesn't add caching because I wasn't entirely sure how to avoid contamination between runs.


mishazharov (approved):
LGTM

PR #471, opened 2021-07-19
More CI improvements

Phantomical:
This fixes the caching setup within the `build-apt` action by using a prebuilt action to do it instead of doing it ourselves. It also adds a debug build which should hopefully be faster for initial smoke tests.


mishazharov (approved):
LGTM

PR #472, opened 2021-07-21
Implement data structure for performing subset queries

Phantomical:
This PR introduces a new data structure called a set-trie. It allows for efficiently checking whether another set is a superset of any set contained within the set-trie itself.

In the following PR this will be used to efficiently check whether a query is a superset of a known unsat core.


mishazharov (approved):
LGTM!

PR #473, opened 2021-07-22
Add `libcxx` and `libcxxabi` linking for tests

mishazharov:
This PR adds `libcxx` and `libcxxabi` as LLVM IR files and links them in for tests if `libc` is enabled. Maybe in the future it would be beneficial for the user to compile it themselves but then they have to have LLVM locally which is a pretty large install size.


Phantomical (commented):
I'm not entirely sure about the approach of embedding the LLVM IR representation of all of libcxx into our repo in readable form. Could we do one of these approaches instead?
- Link it all into one bitcode and embed that in the repo
- Use github actions within the libcxx repo to automatically upload a build artifact and then download that at configure time.

Either way, I would rather it be at most one binary blob that gets dropped into the repo. Embedding the files as textual IR is overkill.

mishazharov:
> I'm not entirely sure about the approach of embedding the LLVM IR representation of all of libcxx into our repo in readable form. 

Yeah I actually agree with this, didn't quite notice the 200kloc change, will update this

libraries/libcxx/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+
+set_directory_properties(PROPERTIES COMPILE_DEFINITIONS "")
+# Disable all warnings since it's not our code.
+set_directory_properties(PROPERTIES COMPILE_OPTIONS "-w")
+
+file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/dummy.c" "")
+
+add_custom_command(
+  OUTPUT libcxx.ll
+  COMMAND curl -L -O `curl -s 'https://api.github.com/repos/insufficiently-caffeinated/libcxx/releases/latest' | jq -r '.assets[0].browser_download_url'`
```
Phantomical, line 11: Can you add that `jq` is required to the readme? It's not guaranteed to be installed on all systems.

libraries/libcxx/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+if (CAFFEINE_ENABLE_LIBC)
+
+set_directory_properties(PROPERTIES COMPILE_DEFINITIONS "")
+# Disable all warnings since it's not our code.
+set_directory_properties(PROPERTIES COMPILE_OPTIONS "-w")
+
+file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/dummy.c" "")
+
+add_custom_command(
```
Phantomical, line 9: Also, maybe you want to replace this with some use of `file(DOWNLOAD ...)` instead of relying on curl? This would help with reducing dependencies.

Phantomical (approved):
Approved but I've left some comments you might want to address.

PR #477, opened 2021-07-22
Small CI cleanups

Phantomical:
This addresses a couple of small CI issues that popped up. It also removes the outdated and slow CI scripts.


mishazharov (approved):
Sweet!

PR #478, opened 2021-07-23
Ensure `clang-tidy` diagnostics use colour

Phantomical:
When building with ninja clang-tidy doesn't think it's outputting to a terminal so it doesn't use any color in the output. This change fixes that by telling it to always use colour.


mishazharov (approved):
LGTM!

PR #480, opened 2021-07-26
`Invoke` instruction implementation

mishazharov:
This PR allows us to symbolically execute c++ code which never throws exceptions.


src/Interpreter/Interpreter.cpp
```diff
@@ -396,28 +398,39 @@ ExecutionResult Interpreter::visitSwitchInst(llvm::SwitchInst& inst) {
...
                               func->getName().str()));
 
-  if (func->empty())
-    return visitExternFunc(call);
+  if (func->empty()) {
+    ExecutionResult res = visitExternFunc(callBase);
+    if (!ctx->stack.empty()) {
+      performInvokeReturn(ctx->stack_top(), callBase);
+    }
+    return res;
```
mishazharov, line 416: This is the cruftiest part of the PR because:
1. Each `Invoke` instruction instance does not necessarily have a matching `Return` since these external functions are handled by Caffeine. As a result, I threw a `performInvokeReturn` at the end here to fix that
2. Maybe I could also template this instead so that we have a compile time check for whether this is an `Invoke` instruction instead of a runtime one

Phantomical: This will break if any builtin function does a non-standard return (e.g. `longjmp`). I think to make this work you need to check that the current instruction has not been changed.

Could you also add a test case for this? Just one where you invoke a function and then `longjmp` away should enough to see whether this breaks.

Phantomical (commented):
This generally looks good to me but I think the `perfomInvokeReturn` logic doesn't work for non-standard control flow. Could you fix that?

src/Interpreter/Interpreter.cpp
```diff
@@ -873,7 +892,22 @@ ExecutionResult Interpreter::visitBuiltinResolve(llvm::CallInst& call) {
     }
   }
 
+  performInvokeReturn(call);
```
mishazharov, line 895: wait this is wrong, the forks also need `performInvokeReturn`

src/Interpreter/Interpreter.cpp
```diff
@@ -830,10 +847,12 @@ ExecutionResult Interpreter::visitFree(llvm::CallInst& call) {
     fork.heaps[ptr.heap()].deallocate(ptr.alloc());
   }
 
+  performInvokeReturn(call);
```
mishazharov, line 850: need to do the same with the forks

src/Interpreter/Interpreter.cpp
Phantomical: You can probably ignore this since `resolve` will never be called by outside functions. Up to you

mishazharov: Already implemented :p

src/Interpreter/Interpreter.cpp
```diff
@@ -396,28 +398,34 @@ ExecutionResult Interpreter::visitSwitchInst(llvm::SwitchInst& inst) {
...
+    return visitIndirectCall(callBase);
 
   CAFFEINE_ASSERT(!func->isIntrinsic(),
                   fmt::format("bad function call '{}': intrinsic function "
                               "calls should be handled by visitIntrinsic",
                               func->getName().str()));
 
   if (func->empty())
-    return visitExternFunc(call);
+    return visitExternFunc(callBase);
```
Phantomical, line 413: Ok so my thought for this was something like this here
```suggestion
  if (func->empty()) {
    auto prev_bb = ctx->stack_top().current_block;
    auto res = visitExternFunc(callBase);

    auto invoke = llvm::dyn_cast<llvm::InvokeInst>(&callBase);
    if (!invoke)
      return res;

    if (res.empty()) {
      if (ctx->stack_top().current_block == prev_bb)
        performInvokeReturn(*ctx, invoke);
    } else {
      for (auto& ctx : res.contexts()) {
        if (ctx.stack_top().current_block == prev_bb)
          performInvokeReturn(ctx, invoke);
      }
    }
```
This way, we don't have to call `performInvokeReturn` for _every single external function_ and it will avoid doing things when the builtin function does bizarre things with control flow.

src/Interpreter/Builtins/setjmp-longjmp.cpp
```diff
@@ -123,19 +123,30 @@ ExecutionResult Interpreter::visitSetjmp(llvm::CallInst& inst) {
...
+
+    if (state.ctx.stack.empty()) {
+      return;
+    }
+
+    if (invoke) {
+      // If the parent is an `Invoke` instruction, a call will always
+      // return to the normal branch
+      state.ctx.stack_top().jump_to(invoke->getNormalDest());
+    }
```
Phantomical, line 141: If you use my suggested change above then you can avoid doing this dance altogether.

src/Interpreter/Builtins/setjmp-longjmp.cpp
```diff
@@ -123,19 +123,30 @@ ExecutionResult Interpreter::visitSetjmp(llvm::CallInst& inst) {
...
   });
   ops.assign(&inst,
              ConstantInt::CreateZero(inst.getType()->getIntegerBitWidth()));
 
   return ops.execute(this);
 }
 
-ExecutionResult Interpreter::visitLongjmp(llvm::CallInst& inst) {
+ExecutionResult Interpreter::visitLongjmp(llvm::CallBase& inst) {
```
Phantomical, line 149: You've accounted for `setjmp` being `invoke`d above but not for `longjmp` jumping to a `setjmp` call that was invoked.

test/run-pass/intrin/longjmp/setjmp-longjmp.cpp
```diff
@@ -0,0 +1,27 @@
+#include "caffeine.h"
+
+#include <csetjmp>
+
+__attribute__((noinline)) void inner_func(jmp_buf buf, int a) {
+  if (a > 5) {
+    throw "oh no a > 5";
+  }
+  longjmp(buf, 11);
```
Phantomical, line 9: Does this generate IR that calls invoke for `longjmp`?

Phantomical (changes requested):
I've left a few comments. A bigger issue is that it looks like you haven't fixed `longjmp` to work with invoke calls. I'm not sure why this doesn't give an error but it should work anyway. 
From my experiments caffeine just kills the context during the `longjmp` call.

Your test case for this doesn't actually use `invoke` for either `setjmp` or `longjmp`. Here's two test cases that should be enough to ensure that it's working.

## `run-fail/intrin/longjmp/invoke-longjmp.cpp`
```llvm
target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128"
target triple = "x86_64-pc-linux-gnu"

%struct.__jmp_buf_tag = type { [8 x i64], i32, %struct.__sigset_t }
%struct.__sigset_t = type { [16 x i64] }

; Here we use invoke to call both setjmp and longjmp
define dso_local void @test() local_unnamed_addr #0  
  personality i8* bitcast (i32 (...)* @__gxx_personality_v0 to i8*)
{
  %buf = alloca %struct.__jmp_buf_tag, align 16
  %res = invoke i32 @_setjmp(%struct.__jmp_buf_tag* nonnull %buf)
    to label %setjmp_normal unwind label %exception

setjmp_normal:
  switch i32 %res, label %failure [
    i32 0, label %setjmp_default
    i32 7, label %setjmp_landing
  ]

setjmp_default:
  invoke void @longjmp(%struct.__jmp_buf_tag* %buf, i32 7)
    to label %failure unwind label %exception

setjmp_landing:
  call void @caffeine_assert(i1 false)
  ret void

exception:
  %ex = landingpad { i8*, i32 } catch i8* null
  br label %failure

failure:
  ret void
}

declare dso_local void @caffeine_assume(i1 zeroext) local_unnamed_addr #1
declare dso_local void @caffeine_assert(i1 zeroext) local_unnamed_addr #1
declare dso_local i32 @_setjmp(%struct.__jmp_buf_tag*) local_unnamed_addr #2
declare dso_local void @longjmp(%struct.__jmp_buf_tag*, i32) local_unnamed_addr #3
declare dso_local i32 @__gxx_personality_v0(...)

attributes #0 = { nounwind uwtable "correctly-rounded-divide-sqrt-fp-math"="false" "disable-tail-calls"="false" "frame-pointer"="none" "less-precise-fpmad"="false" "min-legal-vector-width"="0" "no-infs-fp-math"="false" "no-jump-tables"="false" "no-nans-fp-math"="false" "no-signed-zeros-fp-math"="false" "no-trapping-math"="false" "stack-protector-buffer-size"="8" "target-cpu"="x86-64" "target-features"="+cx8,+fxsr,+mmx,+sse,+sse2,+x87" "unsafe-fp-math"="false" "use-soft-float"="false" }
attributes #1 = { "correctly-rounded-divide-sqrt-fp-math"="false" "disable-tail-calls"="false" "frame-pointer"="none" "less-precise-fpmad"="false" "no-infs-fp-math"="false" "no-nans-fp-math"="false" "no-signed-zeros-fp-math"="false" "no-trapping-math"="false" "stack-protector-buffer-size"="8" "target-cpu"="x86-64" "target-features"="+cx8,+fxsr,+mmx,+sse,+sse2,+x87" "unsafe-fp-math"="false" "use-soft-float"="false" }
attributes #2 = { nounwind returns_twice }
attributes #3 = { noreturn nounwind }

!llvm.module.flags = !{!0}
!llvm.ident = !{!1}

!0 = !{i32 1, !"wchar_size", i32 4}
!1 = !{!"clang version 10.0.0-4ubuntu1 "}
```

## `run-pass/intrin/longjmp/invoke-longjmp.ll`
```llvm
target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128"
target triple = "x86_64-pc-linux-gnu"

%struct.__jmp_buf_tag = type { [8 x i64], i32, %struct.__sigset_t }
%struct.__sigset_t = type { [16 x i64] }

; Here we use invoke to call both setjmp and longjmp
define dso_local void @test() local_unnamed_addr #0  
  personality i8* bitcast (i32 (...)* @__gxx_personality_v0 to i8*)
{
  %buf = alloca %struct.__jmp_buf_tag, align 16
  %res = invoke i32 @_setjmp(%struct.__jmp_buf_tag* nonnull %buf)
    to label %setjmp_normal unwind label %exception

  ; Dead block so that if we are just going to the next instruction then
  ; the test case fails.
dead_block:
  call void @caffeine_assert(i1 false)
  unreachable

setjmp_normal:
  switch i32 %res, label %failure [
    i32 0, label %setjmp_default
    i32 7, label %setjmp_landing
  ]

setjmp_default:
  invoke void @longjmp(%struct.__jmp_buf_tag* %buf, i32 7)
    to label %failure unwind label %exception

exception:
  %ex = landingpad { i8*, i32 } catch i8* null
  br label %failure

failure:
  call void @caffeine_assert(i1 false)
  unreachable

setjmp_landing:
  ret void
}

declare dso_local void @caffeine_assume(i1 zeroext) local_unnamed_addr #1
declare dso_local void @caffeine_assert(i1 zeroext) local_unnamed_addr #1
declare dso_local i32 @_setjmp(%struct.__jmp_buf_tag*) local_unnamed_addr #2
declare dso_local void @longjmp(%struct.__jmp_buf_tag*, i32) local_unnamed_addr #3
declare dso_local i32 @__gxx_personality_v0(...)

attributes #0 = { nounwind uwtable "correctly-rounded-divide-sqrt-fp-math"="false" "disable-tail-calls"="false" "frame-pointer"="none" "less-precise-fpmad"="false" "min-legal-vector-width"="0" "no-infs-fp-math"="false" "no-jump-tables"="false" "no-nans-fp-math"="false" "no-signed-zeros-fp-math"="false" "no-trapping-math"="false" "stack-protector-buffer-size"="8" "target-cpu"="x86-64" "target-features"="+cx8,+fxsr,+mmx,+sse,+sse2,+x87" "unsafe-fp-math"="false" "use-soft-float"="false" }
attributes #1 = { "correctly-rounded-divide-sqrt-fp-math"="false" "disable-tail-calls"="false" "frame-pointer"="none" "less-precise-fpmad"="false" "no-infs-fp-math"="false" "no-nans-fp-math"="false" "no-signed-zeros-fp-math"="false" "no-trapping-math"="false" "stack-protector-buffer-size"="8" "target-cpu"="x86-64" "target-features"="+cx8,+fxsr,+mmx,+sse,+sse2,+x87" "unsafe-fp-math"="false" "use-soft-float"="false" }
attributes #2 = { nounwind returns_twice }
attributes #3 = { noreturn nounwind }

!llvm.module.flags = !{!0}
!llvm.ident = !{!1}

!0 = !{i32 1, !"wchar_size", i32 4}
!1 = !{!"clang version 10.0.0-4ubuntu1 "}
```

test/run-pass/intrin/longjmp/setjmp-longjmp.cpp
Phantomical: Looked into this, it does not.

src/Interpreter/Builtins/setjmp-longjmp.cpp
```diff
@@ -187,6 +187,8 @@ ExecutionResult Interpreter::visitLongjmp(llvm::CallInst& inst) {
     insert_fn(std::move(state));
   });
 
+  std::cout << "tried to longjump" << std::endl;
+
```
Phantomical, line 191: Logging

src/Interpreter/Builtins/setjmp-longjmp.cpp
```diff
@@ -225,12 +227,31 @@ ExecutionResult Interpreter::visitLongjmp(llvm::CallInst& inst) {
...
     }
-    // Don't want to execute _setjmp again
-    ++frame.current;
+
+    // Now we need to check if the caller is an invoke instruction and
+    // go to the default return flow
+    llvm::InvokeInst* invoke =
+        llvm::dyn_cast<llvm::InvokeInst>(&*frame.current);
+    if (invoke) {
+      std::cout << "oh boy we're jumping back to an invoke" << std::endl;
```
Phantomical, line 236: Here

src/Interpreter/Builtins/setjmp-longjmp.cpp
```diff
@@ -225,12 +227,31 @@ ExecutionResult Interpreter::visitLongjmp(llvm::CallInst& inst) {
...
+  std::cout << "did most things related to longjumping" << std::endl;
+
+  ExecutionResult res = ops.execute(this);
+
+  if (res.empty()) {
+    std::cout << "res was empty" << std::endl;
+    std::cout << "res status " << res.status() << std::endl;
+  }
+
+  return res;
```
Phantomical, line 255: Here

test/run-pass/intrin/longjmp/setjmp-longjmp.c
```diff
@@ -1,3 +1,4 @@
+// SKIP TEST
```
Phantomical, line 1: This doesn't need to be skipped since it already passes on master (even though it doesn't work).

Phantomical (approved):
Can you clean up your debug logging? I've pointed out the places that should be cleaned up. Also, there's no need to skip tests that already pass even if they aren't working properly.

LGTM otherwise.

PR #482, opened 2021-07-28
Serializer class and serializer protos

brookedolny:


src/Protos/solver.capnp
```diff
@@ -0,0 +1,16 @@
+@0x9ef46c2155de6bed;
+
+using Cxx = import "/capnp/c++.capnp";
```
brookedolny, line 3: Originally I was going to have the protos here but I ended up adding a separate proto folder in the serializer tool

tools/test-serializer/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+  "${CMAKE_CURRENT_SOURCE_DIR}/Protos/*.capnp"
+)
+
+set(CAPNPC_GEN_DIR "${CMAKE_BINARY_DIR}/gen")
+set(CAPNPC_OUTPUT_DIR "${CAPNPC_GEN_DIR}/test-serializer/")
+file(MAKE_DIRECTORY ${CAPNPC_OUTPUT_DIR})
+
+capnp_generate_cpp(CAPNP_SRCS CAPNP_HDRS "${capnp_schemas}")
+list(APPEND sources ${CAPNP_SRCS})
+message(STATUS "includes: ${includes}")
```
brookedolny, line 28: I'm removing these I just forgot to before I pushed

tools/test-serializer/src/Serializer.cpp
```diff
@@ -0,0 +1,85 @@
...
+void Serializer::write_test_case(int fd) {
+    ::capnp::MallocMessageBuilder message;
+    TestCase::Builder testcase = message.initRoot<TestCase>();
+    ::capnp::List<Symbol>::Builder testcases = testcase.initSymbols(symbols.size());
+
+    for( unsigned int i = 0; i < symbols.size(); i++ ) {
+        Symbol::Builder v = testcases[i];
+        v.setName(symbols[i].name.c_str());
+        auto value = v.initSymbol();
+        const byte * b = reinterpret_cast<const unsigned char *>(symbols[i].bytes);
```
brookedolny, line 27: `byte` is defined as an `unsigned char`

tools/test-serializer/include/Serializer.h
```diff
@@ -0,0 +1,40 @@
...
+namespace caffeine {
+
+namespace serializer {
+
+struct TestCaseValue {
+    std::string name;
+    char * bytes;
+};
+
+class Serializer {
```
brookedolny, line 16: Feedback on structure is welcomed

brookedolny: Like please someone look at this lol

tools/test-serializer/Protos/testcase.capnp
```diff
@@ -0,0 +1,27 @@
+@0xa5a2c19ddf8b8759;
```
Phantomical, line 1: This file seems out-of-date.

tools/test-serializer/src/Serializer.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+void Serializer::save_file() {
+    std::FILE * file = std::fopen(filepath.c_str(), "w");
+    int fd = fileno(file);
+    write_test_case(fd);
+}
+
+void Serializer::load_file() {
+    std::FILE * file = std::fopen(filepath.c_str(), "r");
+    int fd = fileno(file);
```
Phantomical, line 74: Note that `fileno` is not part of standard C. It is part of POSIX and won't work on windows.

Phantomical (changes requested):
Overall I think you have the right idea here :). The prototype design looks pretty good to me and the actual conversion to the capnproto types seems to be good.

With that said I think there's some design/organization issues here. I'll list these out along with my suggestions for fixing them. However, if you think you've got a better approach then feel free to do whatever you think is best.
- First off, this code is not usable from the `caffeine` library. My suggestion is that you move this back into the main library (I realize I said something else when initially coming up with the issue) and then later on _in a different PR_ we can export that subset of the source as library that can be used on its own.
- Second, the `Serializer` class has a bunch of responsibilities that would make it rather difficult to use.
  1. It always writes to a file and it opens that file internally. This means that we can't write the output to anything other than a file (e.g. a database or as part of a larger file format).
  2. It has a whole bunch of responsibilities and it's trying to do too much.
  
  My suggestion here is that you simplify the API to only serialization/deserialization - something that looks like this
  ```cpp
  std::vector<char> serialize_test_case(const std::unordered_map<std::string, std::string>);
  std::unordered_map<std::string, std::string> deserialize_test_case(std::string_view data);
  ```
  so that the users of this code can write the data to wherever they want. Note that `std::string` is fine containing nul chars as long as you construct it using the ptr + size ctor.
  
Basically, my suggestions here are to move stuff back into the caffeine library (under `Serialization`) and simplify the API to the minimum required. Then in a follow up PR we can export just that bit as a library that can be used for a `libcaffeinert` library that allows for running with a concrete test case.

tools/test-serializer/include/Serializer.h
```diff
@@ -0,0 +1,40 @@
...
+    std::string filepath;
+    std::mutex file_mutex;
+    std::vector<TestCaseValue> symbols;
+
+    void write_test_case(int fd);
+    void load_test_case(int fd);
```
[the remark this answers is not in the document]
brookedolny: Done (used VectorOutputStream and ArrayInputStream for the inverse)

tools/test-serializer/include/Serializer.h
brookedolny: No more classes :)

tools/test-serializer/Protos/testcase.capnp
```diff
@@ -0,0 +1,27 @@
...
+enum Type {
+  void @0;
+  int @1;
+  float @2;
+  array @3;
+}
+
+struct Value {
+  union {
+    empty @0 :Void;
```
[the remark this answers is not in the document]
brookedolny: This was used mainly because a union must have at least 2 fields and I didn't have another

tools/test-serializer/Protos/testcase.capnp
```diff
@@ -0,0 +1,27 @@
...
+
+using Cxx = import "/capnp/c++.capnp";
+$Cxx.namespace("caffeine::serializer::protos");
+
+enum Type {
+  void @0;
+  int @1;
+  float @2;
+  array @3;
+}
```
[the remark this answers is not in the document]
brookedolny: Deleted

tools/test-serializer/src/Serializer.cpp
brookedolny: Removed

brookedolny:
Okay, I've updated the Serialize functionality to live in the Serialization folder in Caffeine, and it's just two function calls:

```
std::vector<char> serialize_test_case(const std::unordered_map<std::string, std::string>);
std::unordered_map<std::string, std::string> deserialize_test_case(std::vector<char> data);
```

Phantomical:
@brookedolny I think you said this is ready for review so I've moved it out of draft.

include/caffeine/Serialization/Serialize.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+#include <string>
+// #include <mutex> // TODO remove
+#include <unordered_map>
+#include <vector>
+
+namespace caffeine {
+
+namespace serialize {
```
Phantomical, line 10: You can use a compound namespace declaration to avoid having to indent what's inside.
```suggestion
namespace caffeine::serialize {
```

include/caffeine/Serialization/Serialize.h
```diff
@@ -0,0 +1,19 @@
...
+
+  std::vector<char> serialize_test_case(
+      const std::unordered_map<std::string, std::string>& symbols);
+  std::unordered_map<std::string, std::string>
+  deserialize_test_case(std::vector<char>& data);
```
Phantomical, line 15: I suggest using `std::string_view` for `data` here.  `std::vector<char>` is generally somewhat awkward to use.

Phantomical, line 15: I suggest using `std::string_view` for `data` here.  `std::vector<char>` is generally somewhat awkward to use.

Phantomical (approved):
I've got one small nit WRT to API and another formatting suggestion but otherwise LGTM!

include/caffeine/Serialization/Serialize.h
brookedolny: `std::vector<char>` is used here so that `deserialize_test_case` is the inverse of `serialize_test_case`. Using `std::string` (or `std::string_view`) implies the data is text instead of raw (serialized) byte data which has to be processed.

include/caffeine/Serialization/Serialize.h
brookedolny: I forgot I could do that, thanks

PR #483, opened 2021-07-29
Investigating longjmp behaviour

mishazharov:
It seems that after making `ExecutionResult` continue in the `TransformBuilder`, `longjmp` started failing


mishazharov:
Looks like `longjmp` isn't working properly so I'm closing this

PR #484, opened 2021-08-02
Add `posix_memalign` function

mishazharov:
The `posix_memalign` function is used in c++ exception handling for
allocating new exceptions.


mishazharov:
I think it might actually be cleaner if I refactor this a bit. Specifically I want to:
* Add a `malloc` like function that has an alignment parameter
* Perform the store operations in caffeine builtins to make `visitPosixMemalign`'s implementation cleaner

libraries/builtins/caffeine-builtins.h
```diff
@@ -19,6 +19,9 @@
 // This is a more limited version of malloc that expects size != 0
 void* caffeine_malloc(size_t size);
 void* caffeine_calloc(size_t size);
+// This is a version of malloc that allows users to specify alignment. It expects
+// size != 0 && alignment % 2 == 0 && alignment % sizeof(void *) == 0
+void* caffeine_malloc_align(size_t size, size_t alignment);
```
Phantomical, line 24: Nit: Maybe `caffeine_malloc_aligned` would be a better name here.

libraries/builtins/caffeine-builtins.h
```diff
@@ -19,6 +19,9 @@
 // This is a more limited version of malloc that expects size != 0
 void* caffeine_malloc(size_t size);
```
Phantomical, line 20: You can probably delete this and replace it's uses with a call to `caffeine_malloc_align` with alignment 16

Phantomical (approved):
LGTM - A few small nits.

PR #487, opened 2021-08-04
Fix recursive global resolution stack overflow

mishazharov:
Previously, if a global refered to itself, this would cause a stack
overflow because the `ExprEval` would keep attempting to resolve the
object forever.


include/caffeine/Interpreter/ExprEval.h
```diff
@@ -82,7 +83,7 @@ class ExprEvaluator : public llvm::InstVisitor<ExprEvaluator, LLVMValue> {
   LLVMValue visitFunction(llvm::Function& func);
 
   // Extras
-  OpRef visitGlobalData(llvm::Constant& cnst, unsigned AS);
+  void visitGlobalData(llvm::Constant& cnst, Pointer& ptr, unsigned AS);
```
Phantomical, line 86: Passing the allocation directly makes this cleaner
```suggestion
  void visitGlobalData(llvm::Constant& cnst, Allocation& alloc, unsigned AS);
```

src/Interpreter/ExprEval.cpp
```diff
@@ -371,28 +376,24 @@ LLVMValue ExprEvaluator::visitFunction(llvm::Function& func) {
...
+    return;
 
   LLVMValue value = visit(&constant);
 
-  Allocation alloc{ConstantInt::CreateZero(bitwidth), size, alloc_data,
-                   AllocationKind::Alloca, AllocationPermissions::ReadWrite};
-  alloc.write(ConstantInt::CreateZero(bitwidth), type, value, ctx->heaps,
-              layout);
+  OpRef size = ConstantInt::Create(
+      llvm::APInt(bitwidth, layout.getTypeAllocSize(type).getFixedSize()));
```
Phantomical, line 393: `size` seems to be unused here?

test/regression/CMakeLists.txt
```diff
@@ -13,15 +13,15 @@ function(add_regression_test source)
...
+  if(basename MATCHES "^issue-([0-9]+|misc)\.(pass|fail)\.")
+    if(CMAKE_MATCH_2 STREQUAL "fail")
       set(EXPECTED FAIL)
     endif()
   else()
     message(SEND_ERROR
       " Regression test didn't match naming convention!\n"
       " Expected the start of the test name to match\n"
-      "    issue-[0-9]+\.(pass|fail)\.\n"
+      "    issue-([0-9]+|misc)\.(pass|fail)\.\n"
```
Phantomical, line 24: The correct thing to do here is to open a corresponding issue :)

I've gone ahead and opened #488 for you.

Phantomical (changes requested):
Mostly LGTM. Requesting changes because of the change to the cmake regression test setup - that's there so that people actually go and create a corresponding issue.

src/Interpreter/ExprEval.cpp
```diff
@@ -327,27 +327,33 @@ LLVMValue ExprEvaluator::visitGlobalVariable(llvm::GlobalVariable& global) {
...
+  unsigned alignment = global.getAlignment();
+
   auto alloc = ctx->heaps[global.getAddressSpace()].allocate(
-      array.size(), ConstantInt::Create(llvm::APInt(bitwidth, alignment)), data,
+      size, ConstantInt::Create(llvm::APInt(bitwidth, alignment)), alloc_data,
       AllocationKind::Global, perms, *ctx);
 
-  auto pointer = LLVMValue(
+  auto ptr =
       Pointer(alloc, ConstantInt::Create(llvm::APInt::getNullValue(bitwidth)),
```
Phantomical, line 347: I realize you haven't changed this but using `CreateZero` would be better here.
```suggestion
      Pointer(alloc, ConstantInt::CreateZero(bitwidth),
```

Phantomical (approved):
LGTM! There's one line you may want to clean up that you didn't change in this PR.

PR #490, opened 2021-08-05
Add `pthread_once`

mishazharov:
C++ exception support calls `pthread_once` at some point which means
that we need to support a bare bones implementation.


Phantomical (approved):
LGTM

PR #493, opened 2021-08-11
Add optimization passes to caffeine

mishazharov:
We will need optimization passes in order to generate exception
information for c++. In order to implement this I added more options to
the `ExecutorOptions` struct. This leads to a minor interface change
because I want some of these optimization passes to be mandatory so that
the interpreter can depend on the data generated by these passes.


mishazharov:
After talking a bit about design, we decided to scrap this approach

PR #495, opened 2021-08-25
Rework function pointers

### LINKED ISSUE #485: `longjmp` stops execution with valid jump target
Relevant test cases are `test/run-fail/intrin/call-longjmp.ll` and `test/run-pass/intrin/longjmp/invoke-longjmp.ll`.

The problem is that, when `longjmp` is executed with a valid jump target, caffeine stops executing that context completely. What it should be doing instead is actually performing the longjmp.

Phantomical:
This PR changes the size of the allocations used by function pointers to be the number of instructions within the function. This is a pretty small change that doesn't affect much but it'll allow me to refer to jump targets in a uniform way (the pointer to an instruction is now just function ptr + index of instruction). This will be used when trying to fix #485.

### DIFF
--- a/include/caffeine/Memory/MemHeap.h
+++ b/include/caffeine/Memory/MemHeap.h
@@ -333,6 +333,13 @@ public:
   const MemHeap& operator[](unsigned index) const;
 
   /**
+   * Shortcut function for accessing the function heap. This is equivalent to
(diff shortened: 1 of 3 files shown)


mishazharov (approved):
LGTM!

PR #497, opened 2021-08-25
Fix search for `longjmp` target

### LINKED ISSUE #485: `longjmp` stops execution with valid jump target

Phantomical:
The issue here was pretty simple. All this PR does is do the 1-line fix and re-enable the test cases.

Closes #485


mishazharov (approved):
Nice fix, and darn that missing `!` caused a lot of trouble

Issue #492, opened 2021-08-09
Implement `libunwind` functionality in `caffeine`


mishazharov:
In order to support throwing exceptions in C++ we need to implement some functionality from `libunwind` such as:
* `_Unwind_RaiseException`
* ` _Unwind_GetLanguageSpecificData`
* ` _Unwind_SetGR`
* ` _Unwind_SetIP`
* `_Unwind_GetRegionStart`

Useful reading material:
* https://arxiv.org/pdf/1703.02394.pdf
* https://monkeywritescode.blogspot.com/p/c-exceptions-under-hood.html
* https://itanium-cxx-abi.github.io/cxx-abi/abi.html
* https://itanium-cxx-abi.github.io/cxx-abi/abi-eh.html

Useful code:
* https://divine.fi.muni.cz/download.html see `lart/divine/cppeh.h` and `lart/divine/cpplsda.cpp`

Depends on:
* #501

Update:
:bulb: it might actually be possible to just replace `_Unwind_RaiseException` with a `resume` instruction / resume manually and have everything work automatically. The only prerequisite would be to create a mapping from `std::type_info`s to integers for `@llvm.eh.typeid.for` so that we can resume with the right value. And the value that we resume with is a struct of `{_Unwind_RaiseException's arg0, std::type_info mapping result}`. This is because:
1. `__cxa_begin_catch` basically only wants the `_Unwind_Exception *` which we already have since it's passed to `_Unwind_RaiseException`
2. `__cxa_throw` sets everything else up nicely before calling `_Unwind_RaiseException`, and `_Unwind_RaiseException` is the last thing it does
3. None of the personality functions are expected to change the `_Unwind_Exception` so it should be ready to use in `__cxa_begin_catch` right away

mishazharov:
Closed by https://github.com/insufficiently-caffeinated/caffeine/pull/598 and supporting work

PR #498, opened 2021-08-29
Integrate LSDA generation for `c++` code

mishazharov:
This PR integrates `divine`'s [0] LSDA generation method into `caffeine` in order to support #492. I tested this by dumping a module after the LSDA was generated for it. However, I have not validated that the generated LSDA is correct.

It generates LSDA that looks like this (for the `test` function in the `cpp-vector.cpp` test):
```@test.cpp_lsda = internal constant <{ <{ i8, i8, [4 x i8], i8, [9 x i8] }>, { <{ <{ i32, i32, i32, [4 x i8] }> }>, [2 x i8], {}, [0 x i8] } }> <{ <{ i8, i8, [4 x i8], i8, [9 x i8] }> <{ i8 -1, i8 0, [4 x i8] c"\9C\80\80\00", i8 3, [9 x i8] c"\90\80\80\80\80\80\80\80\00" }>, { <{ <{ i32, i32, i32, [4 x i8] }> }>, [2 x i8], {}, [0 x i8] } { <{ <{ i32, i32, i32, [4 x i8] }> }> <{ <{ i32, i32, i32, [4 x i8] }> <{ i32 trunc (i64 sub nuw (i64 ptrtoint (void (i32)* @test to i64), i64 ptrtoint (void (i32)* @test to i64)) to i32), i32 1, i32 trunc (i64 sub nuw (i64 ptrtoint (i8* blockaddress(@test, %5) to i64), i64 ptrtoint (void (i32)* @test to i64)) to i32), [4 x i8] c"\81\80\80\00" }> }>, [2 x i8] zeroinitializer, {} zeroinitializer, [0 x i8] zeroinitializer } }>```

[0] https://divine.fi.muni.cz/download.html (version 4.4.2)

### CODE REVIEW

File: include/caffeine/Support/Bricks/brick-hash.h [hunk 1]
```diff
@@ -0,0 +1,391 @@
...
+
+static const uint64_t prime_0 = UINT64_C(0xEC99BF0D8372CAAB);
+static const uint64_t prime_1 = UINT64_C(0x82434FE90EDCEF39);
+static const uint64_t prime_2 = UINT64_C(0xD4F06DB99D67BE4B);
+static const uint64_t prime_3 = UINT64_C(0xBD9CACC22C6E9571);
+static const uint64_t prime_4 = UINT64_C(0x9C06FAF4D023E3AB);
+static const uint64_t prime_5 = UINT64_C(0xC060724A8424F345);
+static const uint64_t prime_6 = UINT64_C(0xCB5AF53AE3AAAC31);
+
+[[gnu::always_inline]] static inline uint64_t rot64(uint64_t v, unsigned s) {
```
Phantomical, line 37: The `[[gnu::always_inline]]` attributes aren't portable to MSVC. That might be an issue in the future?

Phantomical (changes requested):
I haven't reviewed the added code beyond skimming through it (there's no way I can review 9300 LOC). 

As a more general concern, could you move the added third party code to a `third_party/divine` subfolder? I think should include pretty much everything here? To explain, there are two reasons for this:
1. We can remove the license modification (which I'm not sure about) from the `LICENSE`  file since it's clear stuff in `third_party` could be under a different license. The license here is compatible but we will probably end up doing something like this again.
2. You can then add a readme that explains where the code came from and how to update it in the future should anything change upstream that we want.

In short, I don't think throwing a bunch of third party code under `Support/X` is a good precedent and we should really just have a `third_party` folder for stuff like this.

Phantomical (approved):
LGTM! Some of the files in `third_party` are still using the caffeine namespace. I don't care too strongly about this but feel free to fix if you want to.

PR #499, opened 2021-09-08
Change update-alternatives script to avoid slaving certain tools

Phantomical:
The script wasn't working correctly on my system. This changes it to make all the tools their own masters.


mishazharov (approved):
Nice refactor

PR #504, opened 2021-09-30
Add external stackframe

mishazharov:
This PR adds external stack frames which will be used to track return and resume values for functions called by external functions. It also adds a utility method to setup and tear down external stack frames before / after a function is called from an external method.

I didn't tie this all together yet because in the `Context` we have a vector of `StackFrame`s. However, since `ExternalStackFrame` subclasses from `StackFrame` we cannot store it in the same vector. I think the solution will be to use smartpointers in the `Context` but didn't want to add such a big change in one PR.

### DIFF
--- a/src/Interpreter/StackFrame.cpp
+++ b/src/Interpreter/StackFrame.cpp
@@ -32,6 +32,8 @@ void StackFrame::insert(llvm::Value* value, const LLVMValue& exprs) {
 
 void StackFrame::set_result(std::optional<LLVMValue> result,
                             std::optional<LLVMValue> resume_value) {
+  CAFFEINE_ASSERT(!result.has_value() || !resume_value.has_value());
+
   auto& caller = *std::prev(current);
 
   if (result.has_value())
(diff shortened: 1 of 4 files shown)


src/Interpreter/StackFrame.cpp
```diff
@@ -47,4 +47,17 @@ void StackFrame::set_result(std::optional<LLVMValue> result,
   }
 }
 
+StackFrame::StackFrame() : frame_id(next_frame_id++) {}
+
+void ExternalStackFrame::set_result(std::optional<LLVMValue> result,
+                                    std::optional<LLVMValue> resume_value) {
+  // It would be pretty weird if both of these were set at the same
+  // time, but who am I to judge
```
Phantomical, line 55: If there are actually not valid semantics that you can think of for having both values set at the same time then there should be an assert here validating that. If we find out in the future that this is not the case then we can just remove the assert, until then it's probably a bug.

Phantomical (approved):
Generally LGTM, left one comment WRT to assertions - approved contingent on that being addressed.

PR #511, opened 2021-10-19
Introduce bazel support for compiling caffeine

Phantomical:
This PR introduces support for building caffeine with bazel. It is a simplified version of #502 which doesn't flesh out the build details needed to actually build and run all the test cases. That will be done in follow-up PRs.

## Details
This PR includes:
- Starlark code to set up all of caffeine's dependencies
- Changes needed to caffeine to work with LLVM 13.0.0
- A script to generate a compilation_commands.json database for the bazel compilation commands database
- A small pile of bazel macros for things that are used within caffeine

I've also setup the bazelrc to perform on-disk caching under `~/.cache/bazel-disk-cache` which should mean that builds should be quick if a source file has been built in the same configuration before.

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -31,6 +31,7 @@ jobs:
       - name: Reset accessed time of files within the cache
         shell: bash
         run: |
+          mkdir -p ~/.cache/bazel-disk-cache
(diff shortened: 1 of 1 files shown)


Phantomical:
I'm pretty sure that the message can just be ignored. It's just a warning about some tests taking too long to execute, but it won't cause any issues for us.

mishazharov (approved):
LGTM, added some nits

PR #524, opened 2021-10-29
Add some test cases for bug related to symbolic malloc

Phantomical:
This PR doesn't include an actual fix but after having made changes for debugging the test case passes now so I'm no longer sure what the issue is.

I'm including the changes here since they seem to fix the issue but I'm not actually sure why this happens.

/stack #523


mishazharov (approved):

mishazharov:
@Phantomical can you please merge this PR please? It might be worth merging the test cases even if the other changes are no good

PR #527, opened 2021-11-02
Fix cmake error message if `clang-tidy` is not found

Phantomical:
Previously this would say that `clang-format` was not found which is wrong and confusing. This fixes the error message to actually say to install clang-tidy.


mishazharov (approved):
yeet

PR #532, opened 2021-11-04
Progress output bar

brookedolny:
This is tested with some of the example tests, and it just prints text that says the number of completed contexts / the number of total contexts every time either number is updated.

I believe there's an issue with thread safety here when using cout for both this progress output and the (printing) failure logger.

### DIFF
--- a/include/caffeine/Interpreter/ContextEvent.h
+++ b/include/caffeine/Interpreter/ContextEvent.h
@@ -45,7 +45,7 @@ protected:
   void update() override;
 
 public:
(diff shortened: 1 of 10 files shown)


Phantomical (commented):
Not sure if you want a full review yet so I've mostly just skimmed it. LMK if you want a full review and I'll go through in depth.

Your approach looks good to me :). I didn't see any glaring issues (although I think you've still go some debug print statements lying around). I think this showing that we really need to have a more general way to listen in to all sorts of interpreter events since the need to do this keeps coming up. That's definitely not something you need to do in this PR and I like the way you've done it here.

PR #540, opened 2021-11-09
Write ExternalFunction interface

Phantomical:
This PR introduces the interface for an external function. @mishazharov I realize you were probably going to do this but I needed it now and figured I'd just add it instead of kludging around it not existing yet and having to fix it again later.

The interface itself is pretty simple: all it has is a `call` function that takes an `InterpreterContext` and the function arguments. That's it. That function then does whatever is needed to actually run the function which may include creating an external stack frame if necessary.

### DIFF
--- a/include/caffeine/Interpreter/ExternalFunction.h
+++ b/include/caffeine/Interpreter/ExternalFunction.h
@@ -1,0 +1,48 @@
+#pragma once
+
+#include "caffeine/ADT/Span.h"
+
+namespace caffeine {
+class InterpreterContext;
+class LLVMValue;
+
+/**
+ * Interface for an external (non-LLVM-IR) function.
+ *
+ * To implement a new external function all that is needed is to derive from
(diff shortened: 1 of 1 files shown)


include/caffeine/Interpreter/ExternalFunction.h
```diff
@@ -0,0 +1,48 @@
...
+namespace caffeine {
+class InterpreterContext;
+class LLVMValue;
+
+/**
+ * Interface for an external (non-LLVM-IR) function.
+ *
+ * To implement a new external function all that is needed is to derive from
+ * this class and then register that instance with the interpreter at startup
+ * time.
```
mishazharov, line 14: Can you please add more documentation about why this is different from an `ExternalStackFrame` and why `ExternalStackFrame` is insufficient (to mirror our irl discussion)?

mishazharov (approved):
LGTM minus one nit

PR #541, opened 2021-11-10
Implement coroutines and add an example implementation

mishazharov:
This PR implements more functionality in `ExternalStackFrame`s up to a point where it's possible to implement external functions using the `ExternalStackFrame`. The `ExternalStackFrame` will be used in order to keep state of external functions which can fork, call other functions, and return results.

Example implementation via `CaffeineAssert`.


src/Interpreter/Interpreter.cpp
```diff
@@ -555,8 +562,19 @@ ExecutionResult Interpreter::visitExternFunc(llvm::CallBase& call) {
...
+    std::vector<LLVMValue> args;
+    for (unsigned int i = 0; i < call.getNumArgOperands(); i++) {
+      args.push_back(interp->load(call.getArgOperand(i)));
+    }
+
+    auto frame = std::make_unique<CaffeineAssertFunc>(std::move(args), func);
+
+    // TODO: Handle forks
+    interp->context().stack.push_back(StackFrame(std::move(frame)));
+    return ExecutionResult::Continue;
```
Phantomical, line 575: This should be returning `ExecutionResult::Migrated` if it's using `InterpreterContext` to manage forking.

src/Interpreter/Interpreter.cpp
```diff
@@ -555,8 +562,19 @@ ExecutionResult Interpreter::visitExternFunc(llvm::CallBase& call) {
   CAFFEINE_ASSERT(func->empty(),
                   "visitExternFunc called with non-external function");
 
-  if (name == "caffeine_assert")
-    return visitAssert(call);
+  if (name == "caffeine_assert") {
+    std::vector<LLVMValue> args;
```
Phantomical, line 566: Could we keep this all within `visitAssert` for now? It will eventually get cleaned up but I don't think it helps to make `visitExternFunc` bigger at the moment.

src/Interpreter/Interpreter.cpp
```diff
@@ -555,8 +562,19 @@ ExecutionResult Interpreter::visitExternFunc(llvm::CallBase& call) {
...
+    for (unsigned int i = 0; i < call.getNumArgOperands(); i++) {
+      args.push_back(interp->load(call.getArgOperand(i)));
+    }
+
+    auto frame = std::make_unique<CaffeineAssertFunc>(std::move(args), func);
+
+    // TODO: Handle forks
+    interp->context().stack.push_back(StackFrame(std::move(frame)));
```
Phantomical, line 574: This should probably be a helper within `InterpreterContext` somewhere since it will be used in a bunch of places.

include/caffeine/Interpreter/StackFrame.h
```diff
@@ -25,22 +28,43 @@ class StackAllocation {
...
+  // Coroutine logic implementation
+  virtual void step(InterpreterContext& context) = 0;
 
 protected:
   void set_result(std::optional<LLVMValue> result,
                   std::optional<LLVMValue> resume_value);
+
   friend class StackFrame;
+
+  uint64_t program_counter = 0;
```
Phantomical, line 67: What is this used for?

The actual implementation of how any individual external stack frame figures out what state is in shouldn't be the responsibility of the parent class (they may not use an integer).

include/caffeine/Interpreter/StackFrame.h
```diff
@@ -25,22 +28,43 @@ class StackAllocation {
...
+  std::vector<LLVMValue> args;
+  llvm::Function* func;
 
   virtual std::unique_ptr<ExternalStackFrame> clone() const = 0;
   virtual ~ExternalStackFrame() = default;
-  ExternalStackFrame(uint64_t frame_id,
-                     std::optional<LLVMValue> result_ = std::nullopt,
-                     std::optional<LLVMValue> resume_value_ = std::nullopt);
+  ExternalStackFrame(std::vector<LLVMValue>&& args,
+                     llvm::Function* func = nullptr);
```
Phantomical, line 56: This should probably not be defaulted? A null function is always invalid.

Phantomical (changes requested):
I've got a bunch of small nits on how things are done along with some bigger concerns about values that don't do anything in `ExternalStackFrame`.

include/caffeine/Interpreter/StackFrame.h
mishazharov: Removed the default but it's actually not invalid to have a null function because an `ExternalFunction` declared in a module can call an `ExternalFunction` that's not declared in a module (which would lead to the null pointer)

include/caffeine/Interpreter/InterpreterContext.h
```diff
@@ -244,6 +244,8 @@ class InterpreterContext {
   void emit_failure(std::string_view message, const Model* model,
                     const Assertion& assertion);
 
+  void call_external_function(std::unique_ptr<ExternalStackFrame>&& frame);
```
Phantomical, line 247: Can you add some docs explaining what this does?

Phantomical (approved):
One last nit but otherwise LGTM

include/caffeine/Interpreter/InterpreterContext.h
mishazharov: :p yep, good call

PR #553, opened 2021-11-26
Linecov

Skrinikov:
Adds Line coverage tracking.

Report format:
```
=============Coverage report==============
-> libraries/builtins/string.c: [5,6]
-> libraries/builtins/make_symbolic.c: [5,6,8,9,10]
-> test/run-pass/branch-false.c: [5,6,8,9,10,14,15]
=========================================
```

Next steps:
* Add coverage ignore files/directories
* Print file with line coverage.
* Branch coverage
* Check how to do it for distributed

### DIFF
--- a/include/caffeine/Interpreter/CaffeineContext.h
+++ b/include/caffeine/Interpreter/CaffeineContext.h
@@ -42,7 +42,7 @@ struct CaffeineOptions {
   /**
    * @brief Determines whether to track line coverage
    */
-  bool run_line_coverage = true;
+  bool run_line_coverage = false;
 
   CaffeineOptions() = default;
 };
--- a/include/caffeine/Interpreter/CaffeineContext.h
+++ b/include/caffeine/Interpreter/CaffeineContext.h
@@ -40,8 +40,8 @@ struct CaffeineOptions {
   uint64_t malloc_alignment = 16;
 
   /**
-  * @brief Determines whether to track line coverage
-  */
+   * @brief Determines whether to track line coverage
+   */
   bool run_line_coverage = true;
 
   CaffeineOptions() = default;
--- a/include/caffeine/Interpreter/CaffeineContext.h
+++ b/include/caffeine/Interpreter/CaffeineContext.h
@@ -39,11 +39,6 @@ struct CaffeineOptions {
    */
   uint64_t malloc_alignment = 16;
 
-  /**
-   * @brief Determines whether to track line coverage
-   */
-  bool run_line_coverage = false;
-
   CaffeineOptions() = default;
 };
 
--- a/include/caffeine/Interpreter/CaffeineContext.h
+++ b/include/caffeine/Interpreter/CaffeineContext.h
@@ -39,6 +39,11 @@ struct CaffeineOptions {
    */
   uint64_t malloc_alignment = 16;
 
+  /**
+  * @brief Determines whether to track line coverage
+  */
+  bool run_line_coverage = true;
+
   CaffeineOptions() = default;
 };
 
@@ -136,7 +141,7 @@ public:
     Builder& with_options(const CaffeineOptions& options);
 
     // Set the coverage counter used by this context.
-    Builder& with_coverage(CoverageTracker& tracker);
+    Builder& with_coverage(std::unique_ptr<CoverageTracker>&& tracker);
 
     Builder& with_default_functions();
     Builder& with_default_intrinsics();
--- a/include/caffeine/Interpreter/CaffeineContext.h
+++ b/include/caffeine/Interpreter/CaffeineContext.h
@@ -58,6 +58,7 @@ private:
   std::unique_ptr<ExecutionContextStore> store_;
   std::unique_ptr<SolverBuilder> builder_;
   std::unique_ptr<FailureLogger> logger_;
+  std::unique_ptr<CoverageTracker> cov_;
   CaffeineOptions options_;
 
 public:
@@ -67,6 +68,7 @@ public:
   ExecutionPolicy* policy() const;
   ExecutionContextStore* store() const;
   FailureLogger* logger() const;
+  CoverageTracker* coverage() const;
   const CaffeineOptions& options() const;
 
   std::shared_ptr<Solver> build_solver() const;
@@ -85,6 +87,7 @@ public:
     std::unique_ptr<ExecutionContextStore> store_;
     std::unique_ptr<SolverBuilder> builder_;
     std::unique_ptr<FailureLogger> logger_;
+    std::unique_ptr<CoverageReporter> cov_;
     CaffeineOptions options_;
 
   public:
@@ -130,6 +133,9 @@ public:
 
     // Set the options used by this context.
     Builder& with_options(const CaffeineOptions& options);
+
+    // Set the coverage counter used by this context.
+    Builder& with_coverage(const CoverageTracker& tracker);
 
     Builder& with_default_functions();
     Builder& with_default_intrinsics();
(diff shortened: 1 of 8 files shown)


include/caffeine/Support/Coverage.h
```diff
@@ -0,0 +1,62 @@
...
+
+public:
+    CoverageTracker() = default;
+    ~CoverageTracker() = default;
+
+    CoverageTracker(const CoverageTracker&) = default;
+    CoverageTracker(CoverageTracker&&) = default;
+
+    // CoverageTracker operator=(const CoverageTracker&) = default;
+    // CoverageTracker operator=(CoverageTracker&&) = default;
```
Phantomical, line 31: You can probably just delete this - it's probably not necessary unless you've got some mix of defaulted and deleted constructors.

I would keep the `CoverageTracker() = default` though since that ensures that the default constructor sticks around even if you add others.

include/caffeine/Support/Coverage.h
```diff
@@ -0,0 +1,62 @@
...
+public:
+    Line() = default;
+    Line(size_t line_num, size_t hits = 1);
+    ~Line() = default;
+
+    Line(const Line&) = default;
+    Line(Line&&) = default;
+
+    // Line operator=(const Line&) = default;
+    // Line operator=(Line&&) = default;
```
Phantomical, line 51: Same comment here WRT defaulted ctors/dtors.

include/caffeine/Support/Coverage.h
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Used to keep track how many times this line has been seen
+     * during execution.
+     */
+    void touch(size_t hits = 1);
+    size_t LineNumber() const;
+    size_t Hits() const;
```
Phantomical, line 59: Nit: these should be either camelCase or snake_case.

include/caffeine/Support/Coverage.h
```diff
@@ -0,0 +1,62 @@
...
+    CoverageTracker() = default;
+    ~CoverageTracker() = default;
+
+    CoverageTracker(const CoverageTracker&) = default;
+    CoverageTracker(CoverageTracker&&) = default;
+
+    // CoverageTracker operator=(const CoverageTracker&) = default;
+    // CoverageTracker operator=(CoverageTracker&&) = default;
+
+    void touch(std::string filename, size_t line);
```
Phantomical, line 33: This should either be `std::string_view`, `const std::string&`, or `llvm::StringRef` depending on whatever is convenient.

tools/caffeine/main.cpp
```diff
@@ -158,10 +159,12 @@ int main(int argc, char** argv) {
     return 2;
   }
 
+  CoverageTracker cov;
   auto caffeine = CaffeineContext::builder()
                       .with_store(std::move(store))
                       .with_logger(std::make_unique<CountingFailureLogger>(
                           std::cout, function))
+                      .with_coverage(cov)
```
Phantomical, line 167:
```suggestion
                      .with_coverage(CoverageTracker())
```

Phantomical (commented):
You've got some formatting problems left and I've left some nits. Lmk if you want me to give it another review :)

You can fix the formatting by either `make format` or `bazel run :format`. Both should (hopefully) be fairly fast.

include/caffeine/Support/Coverage.h
```diff
@@ -0,0 +1,62 @@
...
+class CoverageReport {
+public:
+    CoverageReport();
+    ~CoverageReport();
+};
+
+class CoverageTracker {
+private:
+    std::vector<std::map<size_t, Line>> file_lines;
+    std::map<std::string, size_t> file_to_vec;
```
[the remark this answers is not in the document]
Skrinikov: Unordered map may be useful if we decide to make the reporting smarter but as an initial implementation I will switch to `StringMap`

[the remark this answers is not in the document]
Phantomical: Note that `StringMap` is just a more efficient `std::unordered_map` that gives slightly better ergonomics.

src/Interpreter/Interpreter.cpp
```diff
@@ -64,6 +67,13 @@ void Interpreter::execute() {
   traceblock.close();
 }
 
+void Interpreter::visit(llvm::Instruction& inst) {
+  if (interp->caffeine().options().run_line_coverage) {
+    getInstLine(inst);
+  }
```
Phantomical, line 73: Instead of an option here you could also check whether the context has an instance of the coverage tracker
```suggestion
  if (interp->caffeine().coverage()) {
    getInstLine(inst);
  }
```

src/Interpreter/CaffeineContext.cpp
```diff
@@ -63,6 +64,12 @@ std::shared_ptr<Solver> CaffeineContext::build_solver() const {
   return builder_->build();
 }
 
+CoverageTracker* CaffeineContext::coverage() const {
+  if (cov_)
+    return cov_.get();
+  return nullptr;
```
Phantomical, line 70: This can just be
```suggestion
  return cov_.get();
```

Phantomical (changes requested):
If I approve then this will automatically merge so I'm going to request a few small changes and then I'll approve once you've had a chance to make them. However, other than a few small things the PR LGTM!

One last thing: could you make a command-line option that enables/disables coverage tracking + printing?

include/caffeine/Interpreter/CaffeineContext.h
```diff
@@ -38,6 +39,11 @@ struct CaffeineOptions {
    */
   uint64_t malloc_alignment = 16;
 
+  /**
+   * @brief Determines whether to track line coverage
+   */
+  bool run_line_coverage = false;
+
```
Phantomical, line 46: Remove this since whether or not a tracker is present communicates the same information.

src/Interpreter/CaffeineContext.cpp
```diff
@@ -83,6 +88,10 @@ CaffeineContext Builder::build() {
   if (!logger_)
     throw std::logic_error("No logger provided when building CaffeineContext");
   ctx.logger_ = std::move(logger_);
+  if (ctx.options().run_line_coverage && !cov_)
+    throw std::logic_error(
+        "No coverage tracker provided when building CaffeineContext");
```
Phantomical, line 93: This can also be removed since we're deleting `run_line_coverage`

tools/caffeine/main.cpp
```diff
@@ -158,10 +159,13 @@ int main(int argc, char** argv) {
     return 2;
   }
 
+  std::unique_ptr<CoverageTracker> cov = std::make_unique<CoverageTracker>();
```
Phantomical, line 162: Whether this is constructed or not should be decided by a command-line option. So something like
```cpp
std::unique_ptr<CoverageTracker> cov = nullptr;
if (coverage_enabled)
  cov = std::make_unique<CoverageTracker>();
```

tools/caffeine/main.cpp
```diff
@@ -174,6 +178,10 @@ int main(int argc, char** argv) {
   auto logger = static_cast<CountingFailureLogger*>(caffeine.logger());
   int exitcode = logger->num_failures == 0 ? 0 : 1;
 
+  if (caffeine.options().run_line_coverage) {
```
Phantomical, line 181:
```suggestion
  if (caffeine.coverage()) {
```

Phantomical (commented):
Realized I forgot to do this so I'm leaving the full set of changes WRT to just deciding whether to emit coverage info if there's a coverage tracker in the context.

Phantomical (approved):
LGTM!

PR #555, opened 2021-11-29
Convert `caffeine_assume` to be a proper external function

Phantomical:
This PR does 3 main things:
- The biggest one is the one from the title: convert `caffeine_assume` to be an external function that derives from `ExternalFunction` and is included in the default external function map.
- Beyond that, I've also introduced a test case to verify that invoking `caffeine_assert` works as expected.
- Finally, it includes a small fix to `caffeine_assert` to ensure that when it is invoked it returns to the right location.

/stack #554


Phantomical:
@Mergifyio rebase

mishazharov (approved):

PR #557, opened 2021-11-29
Convert `caffeine_malloc_align` to be an `ExternalFunction` instance

Phantomical:
This is another fairly straightforward one. The rewrite fixes the implementation to validate the signature that `caffeine_malloc_align` is called with instead of causing assertion failures when it doesn't match.

/stack #554


Phantomical:
@Mergifyio rebase

mishazharov (approved):

PR #562, opened 2021-12-03
Add builder functions for all external functions

Phantomical:
As in title, this means that we don't have to add a new header include to `CaffeineContext.cpp` every time we add a new external function. It also means that external functions can be entirely contained within their source files which is also nice.


Skrinikov (approved):
LG

PR #571, opened 2021-12-08
Convert `visitLoadInst` to use `InterpreterContext`

Phantomical:
This one used `TransformBuilder` instead of directly using `ctx` so it got missed in the first pass. Otherwise it's another straightforward conversion. I've taken advantage of some methods within `OperationBuilder` so I've stacked this change on top of that one. In addition, I've also added one more utility method to `InterpreterContext`.

/stack #570


mishazharov (approved):

Phantomical:
@Mergifyio rebase

PR #577, opened 2021-12-19
Rewrite bazel compile-commands generation to use a third-party repo

Phantomical:
Recently, I found https://github.com/hedronvision/bazel-compile-commands-extractor which is an existing solution for building a `compilation-commands.json` database based on processing `bazel aquery`. We had an existing homebrew tool to do this but it never quite worked correctly. By converting this to use a third-party solution we at least get something other people are ensuring will keep working.

I've also deleted the old tool since it's now unused. Furthermore, I've added an `external` symlink that makes bazel autocomplete work as expected.


third_party/hedron/hedron.diff
```diff
@@ -0,0 +1,18 @@
...
+     # Bazel gotcha warning: If you were tempted to use `bazel info execution_root` as the build working directory for compile_commands...search ImplementationReadme.md to learn why that breaks.
+ 
++    visited = set()
++
+     # Dump em to stdout as compile_commands.json entries
+     for files, command in outputs:
+         for file in files:
++            if file in visited:
++                continue
++            visited.add(file)
```
mishazharov, line 14: Is it also possible to upstream this?

mishazharov (approved):
Awesome, LGTM. One question about upstreaming but feel free to merge

third_party/hedron/hedron.diff
Phantomical: Upstream does not want to do this. See https://github.com/hedronvision/bazel-compile-commands-extractor/issues/6. For them the generated `compile_commands.json` is only 250MB while for us it's 5.7GB.

cpsauer: Hey guys :) Creator of that tool, saw this thanks to GitHub backlinks. Puts a big smile on my face that you guys are using it!

I'd be delighted to take changes and work with you guys on this. If the existing behavior requires a patch for you, that's reason enough to get a fix in. Sorry for any miscommunication to the contrary; I'd neglected the optimization because I wasn't sure if it was needed, but sure looks like it is :) I should have just built it up front.

Phantomical: @cpsauer Thanks for writing the tool! I'd be happy to submit a PR with the patch if you want it :)

cpsauer: Realized we probably wanted to do something *slightly* different--but that still accomplished the same goal. So I just knocked it out tonight.

Fixed the file size in https://github.com/hedronvision/bazel-compile-commands-extractor/commit/084957eaa1bf6e2bd031f50b1f5d04c89273103a by only emitting one entry per *header.*

This means that compile_commands.json still might contain multiple entries per source if, for example, it's compiled for multiple platforms or architecture. This is important because that way compile_commands.json still contains all compile commands used to build the project for folks that need that. But it still keeps compile_commands.json.

Merry Christmas to you both! Please lmk (especially) if this doesn't work for you.
-Chris

mishazharov: Hi @cpsauer, thanks for adding a fix! We'll test it soon. Merry Christmas to you as well!

Issue #581, opened 2021-12-23
Clean up `Operation`


Phantomical:
The implementation of `Operation` is currently quite a mess. I'd like to instead clean up the internals so that it makes more sense. The main points of this issue are:
- [ ] Move `Operation` to it's own header so that we can de-clutter `Operation.h`
- [ ] Add an alternative data representation to `Operation` that doesn't use a variant.
- [ ] Convert all public APIs to use the new internal data representation.
- [ ] Remove the old data representation.

This should mean that it's easier to make changes

PR #582, opened 2021-12-23
Move `Operation` to a separate header

Phantomical:
This is the first step of #581. It makes the following changes
- Move `Operation` to a new header file `OperationBase.h`
- Make `Operation` non-copyable and refactor and/or delete all cases that tried to copy an operation instance

Luckily, nothing really copies operation instances so the changes needed to do this were straightforward.


Phantomical:
@Mergifyio rebase

mishazharov (approved):

PR #583, opened 2021-12-23
Clean up `OperationSimplifier` 

Phantomical:
With the changes from #582 it's no longer necessary for `OperationSimplifier` to support both copies and moves. This PR converts it to only do moves and only take non-const `Operation` references.

/stack #582


mishazharov (approved):

Phantomical:
@Mergifyio rebase

PR #584, opened 2021-12-23
Clear `Operation` fields upon move-out

Phantomical:
Right now, `Operation`'s move constructor is basically a copy constructor. This has the potential to hide a number of bugs in how `Operation` instances are used. Currently, there are no issues that this exposes but I'm including it as a safeguard for when I start mucking around in the `Operation` internals.

/stack #583


mishazharov (approved):

Phantomical:
@Mergifyio rebase

PR #585, opened 2021-12-23
Introduce alternative data representation for operation data

Phantomical:
This PR introduces a new class hierarchy for storing the non-operand data of an `Operation`. It's not currently used anywhere but follow-on commits will use it as I convert more operation types to use `OperationData` to store non-operand data.

/stack #584

### DIFF
--- a/include/caffeine/IR/OperationBase.h
+++ b/include/caffeine/IR/OperationBase.h
@@ -203,9 +203,15 @@ protected:
   Type type_;
   Inner inner_;
 
+  std::unique_ptr<OperationData> data_;
+  llvm::SmallVector<OpRef, 4> operands_;
+
   friend llvm::hash_code hash_value(const Operation& op);
 
 protected:
+  Operation(
+    std::unique_ptr<OperationData>&& data,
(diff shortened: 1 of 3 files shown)


include/caffeine/IR/OperationBase.h
```diff
@@ -302,6 +307,145 @@ class Operation : private CopyVTable,
...
+           op->opcode() == Opcode::ConstantArray;
+  }
+
+  llvm::hash_code hash() const override;
+
+private:
+  Symbol symbol_;
+};
+
+class ConstantIntData : public OperationData {
```
mishazharov, line 393: This interface seems really verbose? Do we need a data class for every type of `OpCode`? Is this for efficiency reasons over using some kind of map?

Phantomical: No. We only need a new class if the opcode is storing custom data that's not operands. Note that there's only 3 data classes defined here. Those are the only ones we need, the rest are just ways of storing operands.

mishazharov (approved):

PR #588, opened 2021-12-24
Convert `Constant`, `ConstantInt`, `ConstantFloat`, and `FunctionObject` to use `OperationData`

Phantomical:
This is all the easy ones that don't have any operands. The conversion was fairly straightforward. The one exception is that I needed to fix up the comparison operator for `Operation` so that caching and hashmaps worked as expected.

### DIFF
--- a/src/IR/OperationBase.cpp
+++ b/src/IR/OperationBase.cpp
@@ -15,6 +15,24 @@ std::string_view OperationData::opcode_name(Opcode op) {
   return Operation::opcode_name(op);
 }
 
+bool OperationData::operator==(const OperationData& op) const {
+  if (type_ != op.type_ || opcode_ != op.opcode_)
+    return false;
+
+  if (auto data = llvm::dyn_cast<ConstantData>(this))
+    return data->symbol() == llvm::cast<ConstantData>(op).symbol();
+  if (auto data = llvm::dyn_cast<ConstantIntData>(this))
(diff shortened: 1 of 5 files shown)


src/IR/OperationBase.cpp
```diff
@@ -15,6 +15,24 @@ std::string_view OperationData::opcode_name(Opcode op) {
...
+
+  if (auto data = llvm::dyn_cast<ConstantData>(this))
+    return data->symbol() == llvm::cast<ConstantData>(op).symbol();
+  if (auto data = llvm::dyn_cast<ConstantIntData>(this))
+    return data->value() == llvm::cast<ConstantIntData>(op).value();
+  if (auto data = llvm::dyn_cast<ConstantFloatData>(this))
+    return data->value() == llvm::cast<ConstantFloatData>(op).value();
+  if (auto data = llvm::dyn_cast<FunctionObjectData>(this))
+    return data->function() == llvm::cast<FunctionObjectData>(op).function();
+  return true;
```
mishazharov, line 30: This feels like a bit of a footgun if we add more data types. The implementer needs to remember to add this check as well otherwise it will always return true. Not sure how to solve it but wanted to point it out, maybe:
* if data is null return true
* if not, go through the cases
* if none of the cases match raise an exception

mishazharov (approved):

PR #589, opened 2021-12-25
Update hedron_compile_commands to latest

cpsauer:
Hey guys :)

(Follows up on discussion in https://github.com/insufficiently-caffeinated/caffeine/pull/577)

Removes need for patch. Smaller file size baked in.

Separately, I'd strongly recommend using Renovate to get automatic PRs to stay up to date with this and other libraries!
Also probably worth having refresh_compile_commands more specifically point to your output binary target. 
(More info on both in the [README](https://github.com/hedronvision/bazel-compile-commands-extractor))
I'll leave those to you guys, but here for questions.

Merry Christmas,
Chris


Phantomical:
Thank you for the update! I've tried this out locally and it looks good :)

Merry Christmas as well!

Phantomical (approved):

PR #590, opened 2021-12-27
Add testcase for `strlen` and unbust `libc`

mishazharov:
Previously, `libc` wouldn't link for non c++ files (this might be something I broke in the past while linking in `libcxx`). Additionally I might have found a bug with our memory allocation stuff because a really simple `strlen` testcase fails.


Phantomical:
Do you know what the bug in the memory allocator was? If so, would you also be able to build a minimal test case that reproduces it?

mishazharov:
Yep, actually I'll just split this into about 3 PRs

PR #597, opened 2021-12-29
Fix read write permissions

mishazharov:
Previously we were allocating Read Only memory as Write only memory.
This PR fixes that.

/stack #596


Phantomical (approved):
LGTM with a few comments.

PR #598, opened 2021-12-29
`c++` exception handling: unwinding phase 2 support

mishazharov:
In order to implement exception handling we need to implement two phases:
* phase1, the search phase
* phase2, the exception cleanup phase

This PR implements `phase2` which means that we can throw exceptions and catch them


Phantomical (approved):
LGTM!

PR #599, opened 2021-12-30
Remove temporary file that got committed

Phantomical:
@brookedolny This was added as part of your PR. It seems like it's just the output from some command that was accidentally added. If this is actually meant to be there feel free to reject this PR :)


mishazharov (approved):

mishazharov:
We can also cover this in the meeting

brookedolny:
Crap I actually have no idea how those even got created. Yeah they can be nuked

brookedolny (approved):

PR #600, opened 2021-12-30
Bump Cmake LLVM version to 12.0.0

mishazharov:
Due to some other changes, the Cmake build is currently broken (was
updating the libcxx compilation script)

PR #601, opened 2021-12-30
Implement support for evaluating `GlobalAlias`'

mishazharov:
While working on c++ support I ran into this issue. This adds the fix

/stack #600

### DIFF
--- a/src/Interpreter/Executor.cpp
+++ b/src/Interpreter/Executor.cpp
@@ -1,5 +1,6 @@
 #include "caffeine/Interpreter/Executor.h"
 #include "caffeine/Interpreter/CaffeineContext.h"
+#include "caffeine/Interpreter/ExprEval.h"
 #include "caffeine/Interpreter/Interpreter.h"
 #include "caffeine/Interpreter/Store.h"
 #include "caffeine/Support/UnsupportedOperation.h"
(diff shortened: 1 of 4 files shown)


mishazharov:
Currently failing because of #600

Phantomical:
Not sure why stackbot didn't do anything. Might be because the stacked PR was already merged?

src/Interpreter/Executor.cpp
```diff
@@ -29,6 +30,10 @@ void Executor::run_worker() {
       try {
         Interpreter interp{&ictx};
         interp.execute();
+      } catch (ExprEvaluator::Unevaluatable& ex) {
+        logger->log_failure(nullptr, ctx.value(),
+                            Failure(Assertion(), ex.what()));
+        break;
```
Phantomical, line 36: This should already be caught at a higher level since `ExprEvaluator::Unevaluatable` derives from `UnsupportedOperationException`.

Phantomical (approved):
I've disabled auto-merge so you can address the 1 comment before submitting :)

src/Interpreter/Executor.cpp
mishazharov: It is, but no error is printed, making it hard to debug. Also IIRC there's no where to add error logging in the `ExprEval` implementation

Phantomical: Ah makes sense, LGTM then!

PR #616, opened 2022-01-03
Add `llvm::CallBase` to `ExternalFunction` class

mishazharov:
This commit adds a `llvm::CallBase` to the `call` method because the
`abs` intrinsic needs to inspect the return type. Currently it's not
possible to do this with the current `ExternalFunction` class.


Phantomical (approved):
TBH this was something that I'd run into a bunch of times and really needed to be done.

PR #625, opened 2022-01-06
Implement `llvm.abs.*` intrinsic and fix `ICmpOp`

mishazharov:
This PR fixes a bug in `ICmpOp` where a constructor assumed that the operands are interchangeable and also implements the `llvm.abs.*` intrinsic

### DIFF
--- a/src/Interpreter/Intrinsics/llvm.abs.cpp
+++ b/src/Interpreter/Intrinsics/llvm.abs.cpp
@@ -1,0 +1,44 @@
+#include "caffeine/IR/Value.h"
+#include "caffeine/Interpreter/ExternalFunction.h"
+#include "caffeine/Interpreter/Interpreter.h"
+
+namespace caffeine {
+
+namespace {
+  class AbsIntrinsic : public ExternalFunction {
+  public:
+    void call(llvm::Function* func, InterpreterContext& ctx,
+              Span<LLVMValue> args) const {
+      CAFFEINE_ASSERT(args.size() == 2);
+      CAFFEINE_ASSERT(args[0].is_scalar(),
+                      "Vector operation not supported yet");
+
+      auto should_poison = args[1].scalar().expr();
+      auto val_to_abs = args[0].scalar().expr();
+      auto ret_ty = func->getReturnType();
+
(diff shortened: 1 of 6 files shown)


src/Interpreter/Intrinsics/llvm.abs.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+      auto min_val = ConstantInt::Create(
+          llvm::APInt::getSignedMinValue(ret_ty->getIntegerBitWidth()));
+      auto is_min = ICmpOp::CreateICmpEQ(val_to_abs, min_val);
+
+      Assertion will_succeed = Assertion(BinaryOp::CreateOr(
+          ICmpOp::CreateICmpEQ(should_poison, 0), UnaryOp::CreateNot(is_min)));
+
+      ctx.assert_or_fail(will_succeed,
+                         "Taking absolute number of minimum value");
```
Phantomical, line 28: According to the LLVM spec, poison is only an error in certain cases. We don't really correctly model things - and doing that would be too expensive - so I think this is probably fine for now.

src/Interpreter/Intrinsics/llvm.abs.cpp
```diff
@@ -0,0 +1,44 @@
...
+      auto should_poison = args[1].scalar().expr();
+      auto val_to_abs = args[0].scalar().expr();
+      auto ret_ty = func->getReturnType();
+
+      auto min_val = ConstantInt::Create(
+          llvm::APInt::getSignedMinValue(ret_ty->getIntegerBitWidth()));
+      auto is_min = ICmpOp::CreateICmpEQ(val_to_abs, min_val);
+
+      Assertion will_succeed = Assertion(BinaryOp::CreateOr(
+          ICmpOp::CreateICmpEQ(should_poison, 0), UnaryOp::CreateNot(is_min)));
```
Phantomical, line 25: `should_poison` should always have the type `i1` right? In that case you can just do `ctx.createNot(should_poison)` instead.

Phantomical (commented):
Added, some comments. One is mostly just a but I figured I'd give you a chance to address it.

Phantomical (approved):

PR #634, opened 2022-01-07
Itanium exception handling `phase1` unwinding

mishazharov:
This PR adds `caffeine_unwind_phase1` which takes a `std::type_info *` and returns either `_URC_NO_REASON` if anything can handle the exception or `_URC_END_OF_STACK` if the exception cannot be handled.


src/Interpreter/ExternalFuncs/UnwindPhase1.cpp
```diff
@@ -0,0 +1,239 @@
...
+      auto inst = llvm::cast<llvm::CallBase>(ctx.getCurrentInstruction());
+      if (!inst->getArgOperand(0)->getType()->isPointerTy()) {
+        ctx.fail(
+            "invalid UnwindPhase1Function signature (invalid first argument)");
+        return;
+      }
+
+      ctx.call_external_function(std::make_unique<UnwindPhase1>(
+          std::vector<LLVMValue>(args.begin(), args.end()),
+          inst->getCalledFunction()));
```
Phantomical, line 229: This should be using the function parameter to `call` instead of going through `inst`. The same applies to the type-checking above.

Phantomical (approved):
TBH I don't have the knowledge required here to properly review this PR. I saw one small nit but I don't have the knowledge needed to review the actual unwinding algorithm. If you'd like to walk me through it then that's fine but I'd just suggest that you make sure that you're happy with the test coverage before submitting it.

I've disabled auto-merge so that you can address the nits without having to go back around through me.

In short: LGTM but I don't have the knowledge needed to properly review here.

bazel/bitcode.bzl
```diff
@@ -378,6 +383,14 @@ def caffeine_bitcode_test(should_fail = False, skip = False, **kwargs):
...
 
     name = kwargs["name"]
+    for src in kwargs["srcs"]:
+        _, ext = paths.split_extension(src)
+        if ext in _C_SRC_EXTS:
+            kwargs["deps"] = ["@musl//:libc"]
+            continue
+        if ext in _CPP_SRC_EXTS:
+            kwargs["deps"] = ["@libcxx//:libcxx", "@musl//:libc"]
+            break
```
Phantomical, line 393: Wait, this can't be right. It'll stomp over any existing `deps` that are passed in. Beyond that, it doesn't even work. These dependencies are added through `_builtin_deps` in `bitcode_library`.

This should also be handled within `bitcode_library` since that's where the dependencies are actually added. That would also make sure that there's no compilation differences between test and non-test bitcode files which is actually a rather big issue.

Phantomical (commented):
Can you break out the changes to `bitcode.bzl` into a separate PR? It really doesn't belong as a part of this one.

PR #635, opened 2022-01-08
Add a bazel BUILD file for Z3

Phantomical:
I got fed up with the cmake z3 build not working particularly well with caching or bazel in general. This PR does away with that and just includes a minimal build file for Z3 since it turns out that doing so isn't _that_ hard. It currently exposes all of Z3's internal headers but for the moment that shouldn't be a huge issue.


mishazharov (approved):
Going to trust that this is right because I don't know enough about Z3's build system

PR #636, opened 2022-01-09
Make run-fail tests fail if they hit an unsupported operation

### LINKED ISSUE #633: Better exit code support in tests
Currently if caffeine finds an assertion, or exits due to an `Unsupported operation` being thrown, tests in `run-fail` will pass. However, the behaviour I expected was for tests to pass if an assertion is found, but fail if caffeine crashes

Phantomical:
This fixes #633. It works by explicitly tracking whether a failure was due to an internal abort and overriding the error code if any failure was due to an unsupported operation.

Note that this has uncovered an existing instance of this failure so the tests will fail here. However, the fix for that is already present in #634 so I've decided not to make any changes here. I'll update once #634 is merged.

### DIFF
--- a/tools/caffeine/main.cpp
+++ b/tools/caffeine/main.cpp
@@ -6,15 +6,16 @@
 #include "caffeine/Interpreter/Store.h"
 #include "caffeine/Interpreter/ThreadQueueStore.h"
 #include "caffeine/Solver/LoggingSolver.h"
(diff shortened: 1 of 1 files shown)


mishazharov (approved):

Phantomical:
@Mergifyio rebase

Phantomical:
@Mergifyio rebase

PR #638, opened 2022-01-09
Only link in stdlibs for appropriate languages

mishazharov:
This commit changes how dependencies are linked in for bazel tests.
Currently, we link in libc and libc++ regardless of which language is
actually used. This is undesirable because `.ll` test files don't need
these deps and linking them in can cause builds to fail due to
redefinition of symbols.

### DIFF
--- a/bazel/bitcode.bzl
+++ b/bazel/bitcode.bzl
@@ -15,18 +15,24 @@ BitcodeCcInfo = provider(
     ],
 )
 
-_ALLOWED_SRC_EXTS = [
+_C_SRC_EXTS = [
+    ".c",
+]
+
+_CPP_SRC_EXTS = [
     ".cpp",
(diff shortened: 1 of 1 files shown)


bazel/bitcode.bzl
```diff
@@ -15,18 +15,24 @@ BitcodeCcInfo = provider(
...
+_CPP_SRC_EXTS = [
     ".cpp",
     ".cc",
     ".cxx",
     ".c++",
-    ".c",
-    ".h",
-    ".hh",
     ".hpp",
+    ".hh",
```
Phantomical, line 29: I feel like header files shouldn't be included in this. Can you move them to the list below?

Phantomical (approved):
One comment but otherwise LGTM

PR #641, opened 2022-01-11
Use bazel's remote caching feature in CI

Phantomical:
This PR enables bazel remote caching within CI, along with a number of other changes that are designed to speed up execution of the CI jobs. Here is the summary of what's enabled:
- CI builds now use a bazel cache running on azure via some free credits I have
- I've also enabled `--remote_download_toplevel` which means that builds execute without having to download all the intermediate cached artifacts
- Finally, I've increase the number of downloads that can run in parallel to speed things up even more.

Overall, in getting to this point I've tried a number of different solutions and this one seems to be the best so far.


mishazharov (approved):
Nice fix but I think we're really beating around the bush with this solution. The real issue is that we're building LLVM locally as part of the build for caffeine. Instead we should just build it once in a separate repo and pull in as a release or something since we care about debug symbols and stuff like that. Fundamentally when a user wants to use or develop caffeine, they don't want to also build LLVM.

This would also decrease build times in CI and help developers setting up on new machines (or if we accidentally stomp on our own cache).

Or an option to use the system LLVM would work as well (and that would be used in CI since we're not debugging anything there)

I've created #642 to track this

PR #647, opened 2022-01-17
Implement `llvm.memset.*` intrinsics natively within caffeine

### LINKED ISSUE #646: Implement `llvm.memset.*` intrinsics natively
Running out custom opt plugin over third-party code is inconvenient. We should implement the intrinsics that it supports natively within caffeine since we now have support for that.

This issue is for implementing `llvm.memset.*` intrinsics natively.

Phantomical:
This PR moves the implementation of `llvm.memset.*` into caffeine itself, instead of it being generated by a plugin to `opt`. This simplifies the work needed to set up third-party code.

Closes #646


mishazharov (approved):
Nice change!

PR #650, opened 2022-01-17
Extract `Symbol` out into a separate header

Phantomical:
As in title.

/stack #649


src/IR/Symbol.cpp
```diff
@@ -0,0 +1,40 @@
...
+}
+
+bool Symbol::operator==(const Symbol& symbol) const {
+  return value_ == symbol.value_;
+}
+bool Symbol::operator!=(const Symbol& symbol) const {
+  return !(*this == symbol);
+}
+
+} // namespace caffeine
```
mishazharov, line 40: \n?

mishazharov (approved):
One nit: this isn't formatted

PR #669, opened 2022-01-19
Synchronized output stream

brookedolny:
Currently, this only is used for `std::cout` but could be used for other output streams. `PrintingFailureLogger` and `ContextEventLogger` both use the `sync_ostream_wrapper` for outputting to `std::cout`.


src/Interpreter/PrintingFailureLogger.cpp
```diff
@@ -78,7 +79,9 @@ void PrintingFailureLogger::log_failure(const Model* model, const Context& ctx,
   ss << "Assertion:\n" << failure.check << '\n';
 
   std::unique_lock lock(mtx);
-  *os << ss.str() << std::flush;
+  sync_ostream_wrapper sync(*os);
+  sync << ss.str();
+  // *os << ss.str() << std::flush;
```
Phantomical, line 84: This commented out code should probably just be deleted.

include/caffeine/Support/SyncOStream.h
```diff
@@ -0,0 +1,36 @@
...
+  void flush();
+
+public:
+  sync_ostream_wrapper(std::ostream& o);
+  ~sync_ostream_wrapper();
+
+  std::string get_formatted_string();
+
+  template <typename T>
+  sync_ostream_wrapper& operator<<(T&& s) {
```
Phantomical, line 30: This should take `const T&` since it's not modifying the value.

Phantomical (approved):
LGTM with two nits :)

src/Interpreter/PrintingFailureLogger.cpp
brookedolny: Deleted

include/caffeine/Support/SyncOStream.h
brookedolny: Changed

Phantomical:
Note that you need to rebase this PR since some CI stuff has been updated. Otherwise this should be good to go :)

PR #670, opened 2022-01-20
Add targets to create a release package for caffeine

Phantomical:
This PR uses `rules_pkg` to allow bazel to build a release tarball that installs all the required components to use caffeine from a cmake project. This includes:
- All visible headers from caffeine or any of its dependencies
- The caffeine binary itself
- Caffeine as a shared library which bundles all of its dependencies
- The existing `opt` plugin
- Bitcode files for caffeine-builtins, libc, and libcxx
- `caffeine-config.cmake`
- and any scripts that are meant to be used with vcpkg

In addition, I've also added an install target that allows for installing caffeine to a destination of your choice. Run `bazel run //:install -- --help` to see the flags that can be used here. Beyond that, I've also added an export step to CI so we can get a release tarball from any commit that passed CI if we so choose.

This should be enough to be able to use caffeine from cmake without needing to go through the cmake build system for caffeine at all.


BUILD
```diff
@@ -82,8 +95,8 @@ cc_library(
         "include/caffeine/**/*.def",
     ]),
     copts = WARNING_FLAGS,
-    # strip_include_prefix = "include",
-    includes = ["include"],
+    strip_include_prefix = "include",
+    # includes = ["include"],
```
mishazharov, line 99: Should remove this?

bazel/packaging.bzl
```diff
@@ -0,0 +1,125 @@
...
+        for file in ctx.rule.files.hdrs:
+            path = file.path
+            path = _strip_prefix(path, file.root.path)
+
+            if path.startswith("external/"):
+                path = path.split("/", 2)[2]
+
+            path = _strip_prefix(path, include_prefix)
+
+            # print(file.path, path, include_prefix)
```
mishazharov, line 54: Should remove this

bazel/packaging.bzl
```diff
@@ -0,0 +1,125 @@
...
+        for prefix, replacement in ctx.attr.mappings.items():
+            if path.startswith(prefix):
+                path = "{}/{}".format(replacement, path[len(prefix):].strip("/"))
+
+        path = "{}/{}".format(ctx.attr.prefix, path.strip("/"))
+        # print(path, ":", file)
+
+        outputs[path] = file
+
+    # print(outputs)
```
mishazharov, line 90: Same here

mishazharov (approved):
Nice, LGTM. Have some nits but that's it. Also maybe time to kill the `cmake` build?

Phantomical:
I'd be down with stripping out the majority of the cmake build system once I'm sure people aren't using it day-to-day.

PR #673, opened 2022-01-25
Add utilities for interfacing std::hash and llvm hashing

Phantomical:
All this PR does is add some utilities that can be used to define an implementation of `std::hash` that wraps an existing `llvm::hash_value` implementation.

/stack #672


mishazharov (approved):

Phantomical:
@Mergifyio rebase

PR #678, opened 2022-01-26
Add initial e-graph implementation

Phantomical:
This PR finally introduces the `EGraph` class. Check out the documentation on the class or [the paper](https://arxiv.org/pdf/2004.03082.pdf) for an explanation of how it works. I haven't really tested this code but it's not used anywhere at this point so it should be fine.

/stack #677

### DIFF
--- a/include/caffeine/IR/EGraph.h
+++ b/include/caffeine/IR/EGraph.h
@@ -20,7 +20,7 @@ namespace caffeine {
 class ENode {
 public:
   std::shared_ptr<OperationData> data;
-  llvm::SmallVector<size_t, 2> operands;
+  llvm::SmallVector<size_t, 2> operands = {};
 
   bool operator==(const ENode& node) const;
   bool operator!=(const ENode& node) const;
@@ -31,7 +31,7 @@ public:
 class EClass {
 public:
   std::vector<ENode> nodes;
-  tsl::hopscotch_map<ENode, size_t> parents;
+  tsl::hopscotch_map<ENode, size_t> parents = {};
 
   bool operator==(const EClass& eclass) const;
   bool operator!=(const EClass& eclass) const;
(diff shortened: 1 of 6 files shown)


include/caffeine/IR/EGraph.h
```diff
@@ -0,0 +1,120 @@
...
+  bool operator==(const EClass& eclass) const;
+  bool operator!=(const EClass& eclass) const;
+
+  void merge(EClass& eclass);
+
+  friend llvm::hash_code hash_value(const EClass& node);
+};
+
+/**
+ * An implementation of an E-Graph as described in the egg paper.
```
mishazharov, line 45: ah yes, the egg paper ;)

include/caffeine/Support/Hashing.h
```diff
@@ -22,7 +22,7 @@ struct LLVMHasher {
 
 #define CAFFEINE_DECL_LLVM_HASHER(type)                                        \
   template <>                                                                  \
-  struct ::std::hash<type> {                                                   \
+  struct std::hash<type> {                                                     \
```
mishazharov, line 25: is this new whitespace at the end of the line? might need formatting

mishazharov (approved):
nits and jokes but otherwise lgtm

include/caffeine/IR/EGraph.h
Phantomical: I do link it at the bottom of the comment :P

The paper is actually titled egg though: https://arxiv.org/pdf/2004.03082.pdf

include/caffeine/Support/Hashing.h
Phantomical: New whitespace is due to removing the `::` at the start of the name since that's apparently not valid.

PR #684, opened 2022-01-28
Fix `libunwind` phase1

mishazharov:
This PR fixes a bug in the original phase1 unwinding implementation. Previously, inheritance was not accounted for, but now we use `libcxxabi` to determine whether `std::type_info`s inherit from each other

Changes:
* `__cxxabi_config.h`: required for `private_typeinfo.h`
* `private_typeinfo.h`: required to get `__shim_type_info` which is needed in `caffeine_can_catch`
* `caffeine_can_catch` is able to determine whether `std::type_info` inherit from each other. This is needed to check whether an exception can be caught by the current landing pad
* The rest is just to refactor `UnwindPhase1` to enable it to be reentrant so that we can call `caffeine_can_catch` from it (since it forks).


Phantomical (approved):
LGTM. If we end up having a lot of stuff under builtins we should probably split it up into a couple of smaller libraries.

PR #689, opened 2022-01-31
Integrate EGraphs into the main interpreter path

Phantomical:
As in the title, this puts egraphs into the main interpreter path. To do this I had to make the following changes:
- `Context` now has an additional `egraph` field
- The list of assertions stored in `Context` is now a list of e-class ids. These are converted into an old-style `AssertionList` before making a call to the solver.
- Any expressions that are built through `OperationBuilder` are now directly inserted into the e-graph (with a few exceptions, notably constants and arrays) and it now returns an `Operation` of type `EGraphNode`.
- There's a variety of other smaller changes to add handling of `EGraphNode` where necessary for linking issues.

This PR still has an issue that it causes some `MemHeap::resolve` calls to be doing more times than they should. However, the slowdown for this seems to be minor so it's something that I will address in future follow-up PRs.

/stack #693


mishazharov (approved):
Nice, LGTM with a nice

PR #694, opened 2022-02-04
Compile bitcode tests with `-O3`

Phantomical:
Having to run them in an unoptimized mode makes for really slow tests due to all the load and store queries that must be made to the solver. This resolves the issue by ensuring that they're compiled in the same way they were in the cmake build system.

### DIFF
--- a/test/regression/issue-489.fail.write-to-const-global.c
+++ b/test/regression/issue-489.fail.write-to-const-global.c
@@ -1,5 +1,5 @@
 const int global = 0;
 
-void test() {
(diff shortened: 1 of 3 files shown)


mishazharov (approved):

Phantomical:
Yeah turns out the optimizer saw straight through one of the regression tests and just optimized away the failing bit. I've gone and wrapped that in an `optnone` attribute. Hopefully it doesn't fail now.

mishazharov (approved):

PR #695, opened 2022-02-08
Implement `llvm.eh.typeid.for` intrinsic

mishazharov:
This instrinsic is needed to get the selector values for c++
std::type_infos

### DIFF
--- a/test/run-pass/intrin/typeid_for/BUILD
+++ b/test/run-pass/intrin/typeid_for/BUILD
@@ -1,3 +1,3 @@
 load("//test:tests.bzl", "generate_tests")
 
-generate_tests(skip_files = [])
+generate_tests()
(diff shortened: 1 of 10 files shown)


test/run-pass/intrin/typeid_for/BUILD
```diff
@@ -0,0 +1,3 @@
+load("//test:tests.bzl", "generate_tests")
+
+generate_tests(skip_files = [])
```
Phantomical, line 3: You can drop the `skip_files` argument if it is empty.
```suggestion
generate_tests()
```

Phantomical (changes requested):
The big change I've already mentioned to you in person but I've added a few more small nits as well.

Phantomical (approved):

PR #708, opened 2022-02-14
Integrate e-matching rewrites within the interpreter

Phantomical:
This PR does the final bit of actually applying all the rewrites within caffeine:
- It adds an instance of `EMatcher` to `CaffeineContext` and sets it up with all the default transforms.
- We now call `EGraph::simplify` before running the solver when it is called through `InterpreterContext`

There are also some bugfixes here:
- The associativity rewrite has been rewritten so that it no longer produces an exponential number of rewrites
- I've done some clean-ups so that e-matching always accesses the right e-node and e-class

/stack #696

### DIFF
--- a/src/IR/Symbol.cpp
+++ b/src/IR/Symbol.cpp
@@ -1,4 +1,5 @@
 #include "caffeine/IR/Symbol.h"
+#include <iostream>
 
 namespace caffeine {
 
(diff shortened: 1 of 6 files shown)


src/IR/Symbol.cpp
```diff
@@ -1,4 +1,5 @@
 #include "caffeine/IR/Symbol.h"
+#include <iostream>
```
mishazharov, line 2: Can remove?

mishazharov (approved):
LGTM + nit

src/IR/Symbol.cpp
Phantomical: As it turns out, this is actually a required bugfix. Without it, the `operator <<` implementation for `Symbol` infinitely recurses because it is missing the declaration of `operator <<` for `std::string`.

Issue #637, opened 2022-01-09
Properly handle `SIGINT`


Phantomical:
Currently, when `caffeine` receives `SIGINT` it will ignore it except for the currently running z3 query being interrupted. This is because Z3 registers its own signal handler for `SIGINT`. This behaviour is wrong. We should handle `SIGINT` within caffeine and exit gracefully instead of just eating the signal and effectively doing nothing.

Basic idea for how to do this:
- Register a `SIGINT` handler within main that sets a global flag + calls `interrupt` on all Z3 contexts that are in use.
- Somewhere in the solver stack, have a solver that always returns `Unknown` once the global flag is set but otherwise passes the queries through.
- Do the same thing with a wrapper around the ContextStore. The interrupt handler will likely have to call `shutdown` here as well. (make sure to be careful about acquiring locks here to prevent deadlocks)

PR #717, opened 2022-02-20
Prevent Z3 from intercepting SIGINT within solver calls

Phantomical:
By default, Z3 will intercept a SIGINT signal that occurs within a solver call, swallow it, and return UNKNOWN from the solver. This doesn't at all match with what we want to do on a SIGINT call.

By looking at the Z3 source code, you can see an option "ctrl_c" that supposedly controls whether Z3 will catch SIGINT or not. The previous effort to fix this set the "ctrl_c" parameter on the Z3 context. What the Z3 API does not tell you is that "ctrl_c" is not a valid paramter on the Z3 context and that if you try to set "ctrl_c" on a Z3 context then it will just silently fail to do anything. This obviously meant that the previous attempt at fixing this (by me) did not do anything.

In this attempt, I traced the uses of the "ctrl_c" parameter within the Z3 codebase back to the public API that allows us to set them. As it turns out, the "ctrl_c" option is a _solver_ parameter and not a context or global parameter. I, for one, did not even know there were solver parameters. However, once this is known then the fix is fairly trivial. This commit removes the previous attempt at setting a context parameter and instead just sets the solver parameter after having constructed it.

@Skrinikov this should help somewhat with simplifying what needs to be done for #637

### DIFF
--- a/src/Solver/Z3Solver.cpp
+++ b/src/Solver/Z3Solver.cpp
@@ -192,6 +192,9 @@ SolverResult Z3Solver::resolve(AssertionList& assertions,
   auto block = CAFFEINE_TRACE_SPAN("Z3Solver::resolve");
 
   z3::solver solver = impl->tactic.mk_solver();
+  // We need to disable Ctrl^C at the solver level in order to prevent Z3 from
+  // catching SIGINT when we don't want it to.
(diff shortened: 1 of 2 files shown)


mishazharov (approved):
wow string options + no validation makes me sad

PR #722, opened 2022-02-22
Generalize unwinding

mishazharov:
This PR generalizes unwinding into an abstract class. After the unwind target is found, a subclass can choose what to do with this information. This is because phase 1 and phase 2 unwinding are pretty similar, and only seem to differ in what they do in the end

### DIFF
--- a/include/caffeine/Interpreter/ExternalFuncs/GenericUnwinding.h
+++ b/include/caffeine/Interpreter/ExternalFuncs/GenericUnwinding.h
@@ -1,0 +1,79 @@
+#include "caffeine/Interpreter/ExternalFunction.h"
+#include "caffeine/Interpreter/InterpreterContext.h"
+#include "caffeine/Support/LLVMFmt.h"
+#include <boost/range/algorithm/find_if.hpp>
+#include <fmt/format.h>
+#include <iterator>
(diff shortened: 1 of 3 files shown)


include/caffeine/Interpreter/ExternalFuncs/GenericUnwinding.h
```diff
@@ -0,0 +1,79 @@
+#include "caffeine/Interpreter/ExternalFunction.h"
+#include "caffeine/Interpreter/InterpreterContext.h"
+#include "caffeine/Support/LLVMFmt.h"
+#include <boost/range/algorithm/find_if.hpp>
+#include <fmt/format.h>
+#include <iterator>
+#include <llvm/IR/InstrTypes.h>
+#include <llvm/IR/Instructions.h>
+#include <llvm/IR/Module.h>
```
Phantomical, line 9: Most of these includes aren't needed for the header right? If it's not directly used by the header it should be added to the cpp file.

Phantomical (approved):
It looks like this is pretty much all copy-paste so I haven't reviewed in detail. I have one comment about unnecessary includes in headers but otherwise LGTM.

PR #725, opened 2022-02-22
Fix TypeId lookup

mishazharov:
Stripping bitcasts makes sure we register the global object instead of
some bitcast

### DIFF
--- a/src/Interpreter/TypeidDb.cpp
+++ b/src/Interpreter/TypeidDb.cpp
@@ -1,16 +1,24 @@
 #include <caffeine/Interpreter/TypeidDb.h>
+#include <caffeine/Support/Assert.h>
+
+#include "caffeine/Support/LLVMFmt.h"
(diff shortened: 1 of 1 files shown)


src/Interpreter/TypeidDb.cpp
```diff
@@ -1,16 +1,24 @@
...
 
-  auto val = values.find(exception_ty);
-  if (val != values.end()) {
-    return val->second;
+  {
+    auto val = values.find(exception_ty->stripPointerCasts());
+    if (val != values.end()) {
+      return val->second;
+    }
   }
```
Phantomical, line 19: C++17 has some syntax for this, up to you if you want to use it or not
```
if (auto val = values.find(exception_ty->stripPointerCasts()); val != values.end()) {
  return val->second;
}
```

Phantomical (approved):
LGTM with one suggestion

PR #737, opened 2022-03-06
Assorted bugfixes for `clone_ptr`

Phantomical:
I found a couple of different issues with `clone_ptr` that hadn't been exercised by our codebase yet. This PR goes and fixes them. The specific issues I found were:
- `ClonePointer.h` has no `#pragma once` directive
- Copying a null `clone_ptr` results in a segfault

I've fixed both and added a test case verifying that the second bug is actually fixed.


mishazharov (approved):
Nice catches

PR #742, opened 2022-03-11
Avoid installing headers by default

Phantomical:
The caffeine install package includes a lot of headers by default. However, basically none of the external use-cases for caffeine need this so it doesn't really make sense to install them. This PR splits the `:install` target into `:install` and `:install-full` where the difference between the two is whether they also include the headers or not.

In addition, I've also included a fix to the header-packaging rule so that it correctly handles bazel rules with `include_prefix`

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -71,7 +71,7 @@ jobs:
 
       - name: Run all tests
         run: |
-          bazelisk test --config=ci -- //... \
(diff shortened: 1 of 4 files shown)


mishazharov (commented):
Can you please update the README with how to use the install target as well?

Phantomical:
I've gone and updated the README. @mishazharov could you take another look?

mishazharov (approved):
Noice

PR #743, opened 2022-03-12
Add graceful interrupt handling

### LINKED ISSUE #637: Properly handle `SIGINT`

mishazharov:
This PR implements graceful exit of caffeine if a user sends the SIGINT
signal. This is useful if we run with the `--coverage` flag because then
the line coverage will be dumped at the end of the run

Closes #637

### DIFF
--- a/include/caffeine/Solver/Solver.h
+++ b/include/caffeine/Solver/Solver.h
@@ -238,13 +238,6 @@ public:
   // Add a new solver to the top of the solver stack.
   SolverBuilder& with(const InterFn& func);
 
-  // template <typename T>
-  // SolverBuilder& with() {
-  //   return with([](const std::shared_ptr<Solver>& solver) {
-  //     return std::make_shared<T>(solver);
-  //   });
-  // }
-
   template <typename T, typename... Params>
   SolverBuilder& with(Params... args) {
     auto func = [](const std::shared_ptr<Solver>& solver, Params... args) {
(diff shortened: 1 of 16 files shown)


Phantomical (approved):
LGTM

include/caffeine/Solver/Solver.h
```diff
@@ -231,11 +238,23 @@ class SolverBuilder {
...
-  SolverBuilder& with() {
-    return with([](const std::shared_ptr<Solver>& solver) {
-      return std::make_shared<T>(solver);
-    });
+  // template <typename T>
+  // SolverBuilder& with() {
+  //   return with([](const std::shared_ptr<Solver>& solver) {
+  //     return std::make_shared<T>(solver);
+  //   });
+  // }
```
mishazharov, line 246: TODO: Remove

src/Support/Signal.cpp
```diff
@@ -63,20 +63,36 @@ namespace {
     if (sigactions[sig].sa_handler)
       sigactions[sig].sa_handler(sig);
 
+    if (sig == SIGINT) {
```
mishazharov, line 66: TODO: Disable alarm for SIGINT above as well

PR #754, opened 2022-03-29
Implement support for varargs within caffeine

Phantomical:
This PR implements support for C-style varargs functions on targets using the AMD64 System V ABI. I would have liked to have a universal implementation of varargs however clang does not generate code that makes uses of the `va_arg` LLVM IR instruction so instead we're stuck having to interface with whatever platform-specific code it ends up generating. This also unfortunately means that it's not possible for caffeine to verify whether the code under test is actually using the varargs correctly other than checking whether it goes out of bounds of the allocated argument save area. Nevertheless, this is enough to support pretty much all correct uses of varargs functions which is good enough for me.


mishazharov (approved):
Nice! Not doing a thorough review, but send it

